PR #1, opened 2017-12-11
Renamings following the face-to-face meeting

yrodiere:


backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchIndexCollectorImplMetadata.java
```diff
@@ -28,8 +28,8 @@
 /**
  * @author Yoann Rodiere
  */
-public class ElasticsearchIndexSchemaCollectorImpl<T extends TypeMapping>
-		implements ElasticsearchIndexSchemaElement, IndexSchemaCollector,
+public class ElasticsearchIndexCollectorImplMetadata<T extends TypeMapping>
+		implements ElasticsearchIndexElementMetadata, IndexCollectorMetadata,
```
gsmet: Shouldn't it be `IndexMetadataCollector`?

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchIndexCollectorImplMetadata.java
```diff
@@ -28,8 +28,8 @@
 /**
  * @author Yoann Rodiere
  */
-public class ElasticsearchIndexSchemaCollectorImpl<T extends TypeMapping>
-		implements ElasticsearchIndexSchemaElement, IndexSchemaCollector,
+public class ElasticsearchIndexCollectorImplMetadata<T extends TypeMapping>
```
gsmet: Same as below + the `Impl` part is not in the right place.

engine/src/main/java/org/hibernate/search/v6poc/entity/mapping/building/impl/IndexModelBindingContextImpl.java
```diff
@@ -9,29 +9,29 @@
...
 import org.hibernate.search.v6poc.entity.model.spi.IndexableTypeOrdering;
 import org.hibernate.search.v6poc.entity.model.spi.IndexedTypeIdentifier;
 
 public class IndexModelBindingContextImpl implements IndexModelBindingContext {
 
-	private final IndexSchemaCollector schemaCollector;
+	private final IndexCollectorMetadata schemaCollector;
 	private final IndexModelNestingContextImpl nestingContext;
-	private final EngineHandle engineHandle = new EngineHandle() {
+	private final SearchMetadata searchMetadata = new SearchMetadata() {
```
gsmet: Still find the name a bit blurry but it's definitely better than the old name. Considering it's a blurry notion, the name probably fits perfectly :).

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/spi/FunctionBridge.java
```diff
@@ -23,12 +23,12 @@ default void initialize(BuildContext buildContext) {
 		return null; // Auto-detect the return type and use default encoding
 	}
 
-	R apply(T propertyValue);
+	R toFieldValue(T propertyValue);
```
gsmet: So, I'm definitely +1 on this change. I'm just not totally sold on the `Field` part, considering we sometimes call the entity property a field too. Maybe `toIndexedValue()`/`fromIndexedValue()`?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/spi/IdentifierBridge.java
```diff
@@ -14,9 +14,9 @@
  */
 public interface IdentifierBridge<T> extends AutoCloseable {
 
-	String toString(T id);
+	String toDocumentIdentifier(T id);
```
gsmet: +1

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/processing/impl/FunctionBridgeValueProcessor.java
```diff
@@ -19,21 +19,21 @@
 public class FunctionBridgeValueProcessor<T, R> implements ValueProcessor {
 
 	private final FunctionBridge<T, R> bridge;
-	private final BridgedElementReader<? extends T> bridgedElementReader;
+	private final BridgedElementAccessor<? extends T> bridgedElementAccessor;
```
gsmet: You don't talk about this change in the commit summary. What was the issue with reader? If we can only read the value, I think it's a better name.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/AbstractElasticsearchTypedFieldMetadataContext.java
```diff
@@ -8,14 +8,14 @@
...
+import org.hibernate.search.v6poc.backend.elasticsearch.document.model.ElasticsearchTypedFieldMetadataContext;
 import org.hibernate.search.v6poc.backend.elasticsearch.document.model.impl.esnative.PropertyMapping;
 
 /**
  * @author Yoann Rodiere
  */
-public abstract class AbstractElasticsearchTypedFieldModelContext<T>
-		implements ElasticsearchTypedFieldModelContext<T>, ElasticsearchIndexModelNodeContributor<PropertyMapping> {
+public abstract class AbstractElasticsearchTypedFieldMetadataContext<T>
+		implements ElasticsearchTypedFieldMetadataContext<T>, ElasticsearchIndexModelNodeContributor<PropertyMapping> {
```
gsmet: But you still have `ElasticsearchIndexModelNodeContributor`?

I thought that maybe you were using `Model` for the whole thing and `Metadata` for the separate elements but it doesn't seem to be the case. I think it wouldn't be a bad thing to do so tbh.

yrodiere: Do you have a concrete suggestion? Because it's not easy to find a correct name TBH...

 * `ElasticsearchIndexMetadataContributor` is really ambiguous, because (at least currently) the type doesn't allow to alter the index metadata (which are a different thing in Elasticsearch). Only the "mapping" (in ES wording) / "schema" / "index model". Also
 * `ElasticsearchIndexFieldMetadataContributor` is not correct, because this type is also used to contribute metadata that do not apply to any field in particular. Whether the "_routing" attribute will be mandatory when adding a document to this index, for instance.

With that in mind, "IndexModelNode" really translates what I meant: the implementations contribute to a *part* of the *index model*.

I guess we could go with `ElasticsearchTypeMappingNodeContributor` or something like that if you prefer, relying on the Elasticsearch wording to make the distinction more obvious?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/spi/FunctionBridge.java
yrodiere: Right, will do.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/processing/impl/FunctionBridgeValueProcessor.java
yrodiere: Right, I didn't explain it...
For `IndexFieldAccessor`, the reason is we will need both write (when indexing) and read (when projecting).
For `BridgedElementAccessor`, I must admit I don't remember in details (it's been a while...) but I think the idea was to shield ourselves from a future need to refer to a property in a context where "reading" is not really relevant. There is for example reflection (gathering some metadata through the accessor). We can also reasonably assume that we will need to refer to properties when registering things in `SearchMetadata` for instance, for instance to declare a projection on this or that entity property (as opposed to index field).
I don't have a strong case for Accessor, I'll give you that. It's more that once we go for Reader, then it only makes sense to add reading-related features, and anything else will not fit well.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchIndexCollectorImplMetadata.java
yrodiere: Seems outdated? I renamed it in a later commit.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchIndexCollectorImplMetadata.java
yrodiere: Fixed the Impl part.

yrodiere:
Merging following our discussion. Thanks!

PR #2, opened 2018-01-05
Query DSL changes following the face-to-face meeting

yrodiere:
Two changes:

1. Move to a two-step query definition syntax, separating predicate building from query building
2. Move the Query DSL implementation from the Elasticsearch backend to the engine.
The rationale is the following: we are going to have to add features that are not related to the backend to the query DSL, for instance the ability to query *properties* (as opposed to fields), which would involve field bridges in some way. Thus the DSL should not be implemented directly by the backend, which should only implement some related SPIs.

@gsmet This PR might be of some interest to you. **I strongly advise *against* trying to review the diff**, as it's mainly about moving around code and thus is quite unreadable. But I would appreciate if you could have a look at just `JavaBeanElasticsearchIT.java`, which contains some examples of how to use the query DSL, and tell me if the end result matches your expectations. Is that what you remember we agreed the syntax would look like?


gsmet:
The Travis build failed here: https://travis-ci.org/yrodiere/hibernate-search-6-poc/builds/325484811#L674

gsmet:
I think it's better than what we have now. So let's merge it once the build issues are solved. We can still refine later (not very convinced by the name `PojoSearchTarget` from an end user perspective but we can brainstorm about it later).

yrodiere:
Merged, thanks!

PR #3, opened 2018-01-12
Remove the need for Bridges/Markers to use annotations

yrodiere:
The goal here is to move annotations out of the core of our code, mainly because this has proved a sore point in the past in Hibernate Validator.

This is made possible by reversing the way we work with annotations:

* Previously we had annotations in the core of our code, and "Definition" objects allowed to have an equivalent in programmatic mapping.
* Now we have "BridgeBuilder" objects in the core of our code, and annotations are simply turned into a builder instance.

For the details of this "turning an annotation into a bridge builder" works:

* `org.hibernate.search.v6poc.entity.pojo.bridge.declaration.spi.BridgeMapping` allows to map a bridge annotation to the corresponding builder
* `org.hibernate.search.v6poc.entity.pojo.bridge.mapping.AnnotationBridgeBuilder#initialize` allows to pass on the annotation data to the builder

Incidentally, these changes will also allow us to introduce generic annotations for bridges that don't require any particular parameter: `@Bridge(impl = MyBridge.class)`.

On the bridge implementor side, the main change is that, in order to implement a (parameterized) bridge,  you need to code:

* The bridge (obviously): `class MyBridgeImpl`
* The bridge builder, allowing to set the parameters: `class MyBridgeBuilder implements BridgeBuilder<Bridge>`
* If you want an annotation, the annotation: `@interface MyBridge`

Whereas previously the annotation was mandatory, and what stood for a builder (the "definition" object) was optional.

See `org.hibernate.search.v6poc.JavaBeanElasticsearchIT.MyBridgeImpl` for a simple example, or `GeoPointBridge`/`GeoPointBridgeImpl`/`GeoPointBridgeBuilder` for a more complex one.


mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/builtin/spatial/impl/GeoPointBridgeImpl.java
```diff
@@ -6,45 +6,46 @@
...
 /**
  * @author Yoann Rodiere
  */
-public class GeoPointBridgeImpl implements Bridge<GeoPointBridge> {
+public class GeoPointBridgeImpl implements Bridge {
 
-	private GeoPointBridge parameters;
+	private final String fieldName;
+	private final Store store;
+	private final String markerSet;
```
gsmet: My only comment is that I'm a bit worried that we have to declare these parameters several times. It feels a bit tedious to me.

yrodiere: I understand your concern, but that's not something we can avoid. With this approach, it's actually the annotations that duplicate the parameters that are originally declared in the builder. And obviously annotations must be statically defined, so they need to re-declare the parameters.
Either we think it's acceptable, and then we can merge the PR, or we don't, and then I will have to give up about this.

Some remarks to reassure you:

* Even with the previous solution (the one used in Validator, i.e. a "definition" that builds an annotation instance dynamically), we already need to declare the parameters several times: once in the annotation and once in the "definition" setters. Here we simply have a third declaration for storing the values inside the builder, but hey, that's how bean properties work in Java :)
* If it's declaring "standard" parameters ("store", "fieldName", "analyzer", ...) that worries you, we could simply provide an abstract class implementing all of those, and let people extend it.
* Finally, let's keep in mind that writing a custom bridge, especially one with many parameters, is not something we expect users to do constantly. It's more of a "write once, use everywhere" situation, so a little bit of verbosity when writing the bridge is not the end of the world.

gsmet: The difference with HV is that the `YourConstraintDef` class should be added only if you want to use the programmatic API. You can define a constraint and a validator without defining this additional class.

IMHO, there are 2 different things in my comment:
 - yeah the fact that we need to redeclare the common attributes every time is annoying. I'm not sure we can do better unfortunately, considering the limitations of annotations (you can have a common parent for the API part but not for the annotation unfortunately). Something like `@FieldBridge(bridge = @MyBridge(...), commonParameter1 = ...)` would be ideal. We had the same issue with HV :/. Not sure what we can do to mitigate the issue.
- the quantity of boilerplate code needed to create a bridge. I agree it's not something you will do very often. But I'm wondering if the step will not be too high and if we will have users give up on this.

Agreed that it only concerns advanced bridges.

No real solution here, just wanted to share my concerns from a user POV.

yrodiere:
I'll merge for now, we'll change it later if it turns out to be really awful. Granted, there is a non-negligible amount of boilerplate code required when writing your own Bridge, but I am confident we will be able to make it easier on users:

* Users who just need a one-to-one mapping between a property and a field can use FunctionBridges
* Users who don't need to use programmatic mapping can implement a very simple `AnnotationBridgeBuilder`, without any setter, whose `initialize(Annotation)` method only stores the annotation in a property so it can be passed along to the bridge's constructor in the `build(BuildContext)` method.
* We could provide a variety of solutions to make Bridge implementations less verbose:
** For cases where the Bridge doesn't have any non-standard parameter, we could provide a generic annotation/builder with all the standard parameters (store, analyzer, ...) and an additional parameter to reference the Bridge implementation to use (similar to the "bridge" parameter in the @Field annotation).
** For other cases, we could provide a generic, abstract base class for builders with all the standard parameters. Then boilerplate code would only be about adding whatever non-standard parameters you need, writing the annotation, and writing the code to initialize the builder with the annotation. Part of this builder initialization (the part about standard parameters) could even be automated using reflection.
** If push comes to shove, we could even provide a "compatibility layer" allowing to code a BridgeBuilder and Bridge the old way, with protected methods allowing to build an annotation method in the builder. Then the only boilerplate needed if you only use annotations would be to write a BridgeBuilder extending our "compatibility layer" abstract class, and write the `build(BuildContext)` method in this bridge builder (which essentially would just call the bridge's constructor and pass the BuildContext and annotation to it).

PR #4, opened 2018-01-22
Update the object field mapping/indexing API and implement it for Elasticsearch

yrodiere:
The indexing APIs now allow fine control over object nesting, which is a prerequisite to allowing structured searches ("nested" field type in Elasticsearch, "nested documents" with Lucene).

Here is an example of how a bridge can take advantage of the API:

```java
@Override
public void contribute(IndexSchemaElement indexSchemaElement, PojoModelElement bridgedPojoModelElement,
		SearchModel searchModel) {
	sourceAccessor = bridgedPojoModelElement.createAccessor( List.class );
	IndexSchemaObjectField objectField = indexSchemaElement.objectField( "object" );
	objectFieldAccessor = objectField.createAccessor(); // This declares an object field that can be used later
	textFieldAccessor = objectField.field( "text" ).asString().createAccessor();
}
@Override
public void write(DocumentElement target, PojoElement source) {
	List<String> sourceValue = sourceAccessor.read( source );
	if ( sourceValue != null ) {
		for ( String item : sourceValue ) {
			if ( item != null ) {
				String[] subItems = item.split( "," ); // Simulate a multi-valued input
				// Below we add an object at path "object"
				DocumentElement object = objectFieldAccessor.add( target );
				for ( String subItem : subItems ) {
					// Below we add a value to the field "object.text"
					textFieldAccessor.write( object, subItem );
				}
			}
			else {
				objectFieldAccessor.addMissing( target );
			}
		}
	}
}
```

The "accessor" syntax is admittedly a bit weird... I wonder if we shouldn't (in a follow-up PR) switch to a "reference" syntax, i.e. instead of `fieldAccessor.write(documentElement, value)`, do something like `documentElement.add(fieldReference, value)`. We would have to validate the name again though, because I don't think "accessor" is a good name (no more than "writer") if we use the object that way.

This other syntax would look that way:

```java
@Override
public void contribute(IndexSchemaElement indexSchemaElement, PojoModelElement bridgedPojoModelElement,
		SearchModel searchModel) {
	sourceReference = bridgedPojoModelElement.createReference( List.class );
	IndexSchemaObjectField objectField = indexSchemaElement.objectField( "object" );
	objectFieldReference = objectField.createReference(); // This declares an object field that can be used later
	textFieldReference = objectField.field( "text" ).asString().createReference();
}
@Override
public void write(DocumentElement target, PojoElement source) {
	List<String> sourceValue = source.get( sourceReference );
	if ( sourceValue != null ) {
		for ( String item : sourceValue ) {
			if ( item != null ) {
				String[] subItems = item.split( "," ); // Simulate a multi-valued input
				// Below we add an object at path "object"
				DocumentElement object = target.add( objectFieldReference );
				for ( String subItem : subItems ) {
					// Below we add a value to the field "object.text"
					object.add( textFieldReference, subItem );
				}
			}
			else {
				target.addMissing( objectFieldReference );
			}
		}
	}
}
```


gsmet:
Let's merge this one.

yrodiere:
Merged, thanks!

PR #5, opened 2018-01-23
Rename Bridge.contribute() to bind()

yrodiere:
Following my discussion with Sanne.

The name "bind" is admittedly very generic, but it has the advantage of not being misleading, as it does not dismiss any of the purposes of this method. "contribute" would just make it feel like this method is only about outputing information to the schemas, whereas it is also about initializing the bridge so it can actually do something in the `write()` method.

Hopefully a good javadoc will make it clear exactly what we expect from users in this `bind()` method and will offset the generic nature of "bind".

### DIFF
--- a/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/Bridge.java
+++ b/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/Bridge.java
@@ -17,7 +17,7 @@ import org.hibernate.search.v6poc.entity.pojo.model.PojoElement;
  */
 public interface Bridge extends AutoCloseable {
 
-	void contribute(IndexSchemaElement indexSchemaElement, PojoModelElement bridgedPojoModelElement,
+	void bind(IndexSchemaElement indexSchemaElement, PojoModelElement bridgedPojoModelElement,
 			SearchModel searchModel);
 
 	void write(DocumentElement target, PojoElement source);
--- a/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/Bridge.java
+++ b/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/Bridge.java
@@ -9,19 +9,61 @@ package org.hibernate.search.v6poc.entity.pojo.bridge;
 import org.hibernate.search.v6poc.backend.document.DocumentElement;
 import org.hibernate.search.v6poc.backend.document.model.IndexSchemaElement;
 import org.hibernate.search.v6poc.entity.model.SearchModel;
-import org.hibernate.search.v6poc.entity.pojo.model.PojoModelElement;
 import org.hibernate.search.v6poc.entity.pojo.model.PojoElement;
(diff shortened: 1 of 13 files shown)


mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/FunctionBridge.java
```diff
@@ -10,23 +10,58 @@
...
+ *
  * @author Yoann Rodiere
  */
 public interface FunctionBridge<T, R> extends AutoCloseable {
 
+	/**
+	 * Bind this bridge instance to the given index field model.
+	 * <p>
+	 * This method is called exactly once for each bridge instance, before any other method.
+	 * It allows the bridge declare its expectations regarding the index field (type, storage options, ...).
```
gsmet: *to* declare?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/FunctionBridge.java
```diff
@@ -10,23 +10,58 @@
...
+	 */
 	R toIndexedValue(T propertyValue);
 
-	/*
-	 * TODO use this method when projecting.
+	/**
+	 * Transform the given indexed field value back to the value of the POJO property,
+	 * or to any implementation-defined value to be returned in projections on the POJO property.
+	 * <p>
+	 * For instance, a {@link FunctionBridge} indexing JPA entities by putting their identifier in a field
```
gsmet: Maybe {@code } would be better, considering we already are in `FunctionBridge`?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/bridge/Bridge.java
```diff
@@ -6,22 +6,64 @@
...
 
 /**
+ * A bridge between an element of the POJO model and an element of the index schema.
+ * <p>
+ * The {@link Bridge} interface is a more powerful version of {@link FunctionBridge},
+ * in which a property can be mapped to more than one field, in particular.
+ *
  * @author Yoann Rodiere
  */
 public interface Bridge extends AutoCloseable {
```
gsmet: Didn't we talk about renaming this one to `PojoBridge`?

yrodiere: It was about splitting it into two classes, `PojoBridge` (or `PojoClassBridge`, I don't remember) and `PojoPropertyBridge`, if I remember correctly. Yes, we talked about it, but I still have to work on it.

gsmet: OK, remembered the `Pojo` part :). Let's do it later then.

yrodiere:
@gsmet I addressed your comments and rebased.

gsmet:
Let's merge it then!

yrodiere:
Merged!

PR #6, opened 2018-01-24
Add API for "nested" storage of object fields (mapping and search) and Elasticsearch implementation

yrodiere:


backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchIndexModel.java
```diff
@@ -18,11 +18,22 @@
 
 	private final String indexName;
 	private final TypeMapping mapping;
+	private final Map<String, ElasticsearchObjectNodeModel> objectFieldModels = new HashMap<>();
 	private final Map<String, ElasticsearchFieldModel> fieldModels = new HashMap<>();
 
 	public ElasticsearchIndexModel(String indexName, ElasticsearchRootIndexSchemaCollectorImpl collector) {
 		this.indexName = indexName;
-		this.mapping = collector.contribute( fieldModels::put );
+		this.mapping = collector.contribute( new ElasticsearchIndexModelCollector() {
```
gsmet: Isn't it an ElasticsearchIndexSchemaCollector?

I thought we decided that Model = Schema + other things. As we add elements to the root IndexSchemaCollector, I would have expected Schema in there.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/AbstractElasticsearchIndexSchemaCollector.java
```diff
@@ -36,9 +37,10 @@ public String toString() {
 	public abstract ElasticsearchIndexSchemaElement withContext(IndexSchemaNestingContext context);
 
 	@Override
-	public ObjectFieldIndexSchemaCollector objectField(String relativeName) {
+	public ObjectFieldIndexSchemaCollector objectField(String relativeName, ObjectFieldStorage storage) {
 		IndexSchemaObjectPropertyNodeBuilder nestedNodeBuilder =
```
gsmet: The variable name should probably be changed.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchIndexSchemaElementImpl.java
```diff
@@ -54,13 +55,14 @@ public FieldModelContext field(String relativeName) {
...
 	@Override
-	public ElasticsearchIndexSchemaObjectField objectField(String relativeName) {
+	public ElasticsearchIndexSchemaObjectField objectField(String relativeName, ObjectFieldStorage storage) {
 		return nestingContext.nest(
 				relativeName,
 				// If the field is included
 				(prefixedName, filter) -> {
 					IndexSchemaObjectPropertyNodeBuilder nestedNodeBuilder =
 							new IndexSchemaObjectPropertyNodeBuilder( nodeBuilder.getAbsolutePath(), prefixedName );
+					nestedNodeBuilder.setStorage( storage );
```
gsmet: Same here about the variable name.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchObjectNodeModel.java
```diff
@@ -6,12 +6,15 @@
...
 /**
  * @author Yoann Rodiere
  */
 public class ElasticsearchObjectNodeModel {
 
-	private static final ElasticsearchObjectNodeModel ROOT = new ElasticsearchObjectNodeModel( null, null );
+	private static final ElasticsearchObjectNodeModel ROOT =
+			new ElasticsearchObjectNodeModel( null, null, null );
 
 	public static ElasticsearchObjectNodeModel root() {
```
gsmet: I didn't check locally as I have a lot of things going on in my tree but is this still useful now that you have differentiated the root index schema thing?

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -57,5 +57,11 @@ SearchException conflictingFieldFormattersForSearch(String absoluteFieldPath,
 			+ " Given predicate was: '%1$s'" )
 	SearchException cannotMixElasticsearchSearchQueryWithOtherPredicates(SearchPredicate predicate);
 
+	@Message(id = 9, value = "Field '%1$s' is not an object field in index '%2$s'." )
+	SearchException nonObjectFieldForNestedQuery(String absoluteFieldPath, String indexName);
```
gsmet: So it's probably just me but usually I try to be consistent with having the more general notions first, even if it doesn't fit the order of the message. This way it's all consistent.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/SearchPredicateFactoryImpl.java
```diff
@@ -72,14 +80,35 @@ else if ( !selectedFormatter.equals( formatter ) ) {
...
 		}
 		return selectedFormatter;
 	}
 
+	private void checkNestedField(String absoluteFieldPath) {
+		boolean found = false;
+
+		for ( ElasticsearchIndexModel indexModel : indexModels ) {
+			ElasticsearchObjectNodeModel nodeModel =
+					indexModel.getObjectNodeModel( absoluteFieldPath );
```
gsmet: From an API POV, I would have thought of:
```
indexModel.getNodeModel( absoluteFieldPath ).isObject();
```
(or maybe even `isNested()`)

Wondering if it's really worth it to have 2 separate maps for them.

engine/src/main/java/org/hibernate/search/v6poc/backend/document/model/ObjectFieldStorage.java
```diff
@@ -0,0 +1,57 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.v6poc.backend.document.model;
+
+public enum ObjectFieldStorage {
+
+	/**
+	 * Use the backend-specific default.
+	 */
+	DEFAULT,
```
gsmet: Do we really expect to have different behavior per backend? I mean it would really change the results so I'm not sure it's such a good idea (we can revisit that later though).

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/AbstractElasticsearchIndexSchemaCollector.java
yrodiere: Will do.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchIndexModel.java
yrodiere: Ok, will do. I expect we'll have to rename it to *Model* ultimately, but then we'll have to rename some SPIs too, so let's be consistent in the meantime.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchIndexSchemaElementImpl.java
yrodiere: Will do.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/impl/ElasticsearchObjectNodeModel.java
yrodiere: Yes it is. We need to pass an ElasticsearchObjectNodeModel to contributors of the properties of the root node (non-embedded fields).

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/logging/impl/Log.java
yrodiere: So, `nestedQueryNonObjectField`?

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/SearchPredicateFactoryImpl.java
yrodiere: The thing is, `ElasticsearchObjectNodeModel` and `ElasticsearchFieldModel` have very few in common. This `isObject` method, and maybe a `getParent` method, but that's all.
If I did what you're suggesting, `getNodeModel` would have to return some interface with basically no useful method, and I would have to cast the interface somehow. Not sure it would be much cleaner...
I'll think it over, maybe I can find some nicer way to do it than this multiple-for garbage...
Also, this is not API. Not even SPI.

engine/src/main/java/org/hibernate/search/v6poc/backend/document/model/ObjectFieldStorage.java
yrodiere: It doesn't change the results, it only changes whether users can use "nested" queries or not. Without a "nested" query, the "nested" field behaves exactly as an "object" field.
FWIW, the main use I saw for this was the ability for users to specify a default for storage in multi-valued `@IndexedEmbedded`: some people may always want to use "NESTED" storage.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/SearchPredicateFactoryImpl.java
gsmet: Yeah, I was not using API as in API/SPI.

I hear you. It just feels weird.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/logging/impl/Log.java
gsmet: Ah no sorry if I wasn't clear but I was talking about the order of the parameters.

engine/src/main/java/org/hibernate/search/v6poc/backend/document/model/ObjectFieldStorage.java
gsmet: > It doesn't change the results

Well, changing the storage format does change the search results (see your example with John Doe / Jane Smith).

But the main use you have in mind convinces me it's a good idea: it's a nice thing to be able to override the default at a global level.

yrodiere:
@gsmet I added a commit about the nodeModel thing, could you have a look?
The exception nesting is admittedly a bit weird, but I think it's good enough until we add the concept of error collector mentioned here: https://docs.google.com/document/d/16PAa__LsxyLZcbW3q1MvgyIznh4ZnCYLupbNMXKMRds/edit#heading=h.s83333f2qkde
Note: if you disagree with naming, let's discuss that on HipChat, it'll be easier ;)

yrodiere:
Added some more renamings and merged. Thanks!

PR #7, opened 2018-01-26
 Add annotation mapping support

yrodiere:
*WARNING*: This PR is based on #6 , which should be merged first.


mapper-javabean/src/main/java/org/hibernate/search/v6poc/entity/javabean/model/impl/JavaBeanTypeModel.java
```diff
@@ -10,39 +10,69 @@
...
 	private final Class<T> type;
 	private final BeanInfo beanInfo;
+	private final BeanInfo declaredBeanInfo;
 
-	public JavaBeanTypeModel(Class<T> type) throws IntrospectionException {
+	JavaBeanTypeModel(JavaBeanIntrospector introspector, Class<T> type) throws IntrospectionException {
+		this.introspector = introspector;
 		this.type = type;
 		this.beanInfo = Introspector.getBeanInfo( type );
+		this.declaredBeanInfo = Introspector.getBeanInfo( type, type.getSuperclass() );
```
gsmet: I'm a bit surprised you're using a static call here.

mapper-javabean/src/main/java/org/hibernate/search/v6poc/entity/javabean/model/impl/JavaBeanTypeModel.java
```diff
@@ -10,39 +10,69 @@
...
+	public Stream<? extends Annotation> getAnnotationsByMetaAnnotationType(Class<? extends Annotation> metaAnnotationType) {
+		return introspector.getAnnotationsByMetaAnnotationType( type, metaAnnotationType );
+	}
+
 	@Override
 	public PropertyModel<?> getProperty(String propertyName) {
 		try {
 			String normalizedName = Introspector.decapitalize( propertyName );
 			PropertyDescriptor propertyDescriptor = getPropertyDescriptor( normalizedName );
-			return new JavaBeanPropertyModel<>( this, propertyDescriptor.getPropertyType(), propertyDescriptor );
```
gsmet: OK, so, it's coming a bit out of the blue but seeing `JavaBeanPropertyModel` just made me think about it: we had several requests for HV of people wanting us to support different models (especially the generation framework generating `property()` as a getter for property).

See this one: https://hibernate.atlassian.net/browse/HV-1363 .

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/logging/impl/Log.java
```diff
@@ -9,35 +9,42 @@
...
 import org.jboss.logging.BasicLogger;
-import org.jboss.logging.annotations.Cause;
 import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageLogger;
 
 @MessageLogger(projectCode = "HSEARCH-POJO")
 public interface Log extends BasicLogger {
 
-	@Message(id = 1, value = "Unable to create annotation for definition of type %1$s")
-	SearchException unableToCreateAnnotationForDefinition(Class<? extends Annotation> annotationType, @Cause Exception e);
```
gsmet: It's really not worth it to renumber the other methods when removing one. It's not a big deal if there are a few gaps.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/PojoMappingContributor.java
```diff
@@ -6,8 +6,6 @@
  */
 package org.hibernate.search.v6poc.entity.pojo.mapping;
 
-import java.util.Set;
```
gsmet: Don't we have a checkstyle rule for that? Or we haven't enabled checkstyle yet?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
```diff
@@ -0,0 +1,278 @@
...
+		}
+
+		String[] includePathsArray = annotation.includePaths();
+		Set<String> cleanedUpIncludePaths;
+		if ( includePathsArray.length > 0 ) {
+			cleanedUpIncludePaths = new HashSet<>();
+			Collections.addAll( cleanedUpIncludePaths, includePathsArray );
+		}
+		else {
+			cleanedUpIncludePaths = null;
```
gsmet: Is it such a good idea to use `null` here? Having an empty set seems to make more sense to me but maybe I'm missing something.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
```diff
@@ -0,0 +1,278 @@
...
+	private <A extends Annotation> AnnotationMarkerBuilder<A> createMarkerBuilder(A annotation) {
+		MarkerMapping markerMapping = annotation.annotationType().getAnnotation( MarkerMapping.class );
+		MarkerMappingBuilderReference builderReferenceAnnotation = markerMapping.builder();
+		BeanReference builderReference =
+				toBeanReference(
+						builderReferenceAnnotation.name(),
+						builderReferenceAnnotation.type(),
+						MarkerMappingBuilderReference.UndefinedImplementationType.class
+				)
+						.orElseThrow( () -> log.missingBuilderReferenceInBridgeMapping( annotation.annotationType() ) );
```
gsmet: the indentation is a bit weird here.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
```diff
@@ -0,0 +1,278 @@
...
+
+	private BridgeBuilder<? extends IdentifierBridge<?>> createIdentifierBridgeBuilder(
+			DocumentId annotation, PropertyModel<?> propertyModel) {
+		IdentifierBridgeBeanReference bridgeReferenceAnnotation = annotation.identifierBridge();
+		Optional<BeanReference> bridgeReference = toBeanReference(
+				bridgeReferenceAnnotation.name(),
+				bridgeReferenceAnnotation.type(),
+				IdentifierBridgeBeanReference.UndefinedImplementationType.class
+		);
+		IdentifierBridgeBuilderBeanReference builderReferenceAnnotation = annotation.identifierBridgeBuilder();
```
gsmet: Maybe rename the variables `bridgeBuilder...`?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
```diff
@@ -0,0 +1,278 @@
...
+					new BeanResolverBridgeBuilder( IdentifierBridge.class, bridgeReference.get() );
+			return castedBuilder;
+		}
+		else if ( builderReference.isPresent() ) {
+			// TODO check generic parameters of builder.getClass() somehow, maybe in a similar way to what we do in FunctionBridgeUtil
+			return beanResolver.resolve( builderReference.get(), BridgeBuilder.class );
+		}
+		else {
+			// The bridge will be auto-detected from the property type
+			return null;
```
gsmet: Wouldn't using an `Optional` make sense here? As `null` seems to be a valid use case we need to take care of.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
```diff
@@ -0,0 +1,278 @@
...
+
+	private BridgeBuilder<? extends FunctionBridge<?, ?>> createFunctionBridgeBuilder(
+			Field annotation, PropertyModel<?> propertyModel) {
+		FunctionBridgeBeanReference bridgeReferenceAnnotation = annotation.functionBridge();
+		Optional<BeanReference> bridgeReference = toBeanReference(
+				bridgeReferenceAnnotation.name(),
+				bridgeReferenceAnnotation.type(),
+				FunctionBridgeBeanReference.UndefinedImplementationType.class
+		);
+		FunctionBridgeBuilderBeanReference builderReferenceAnnotation = annotation.functionBridgeBuilder();
```
gsmet: Same remark here about the variable names.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
```diff
@@ -0,0 +1,278 @@
...
+					new BeanResolverBridgeBuilder( FunctionBridge.class, bridgeReference.get() );
+			return castedBuilder;
+		}
+		else if ( builderReference.isPresent() ) {
+			// TODO check generic parameters of builder.getClass() somehow, maybe in a similar way to what we do in FunctionBridgeUtil
+			return beanResolver.resolve( builderReference.get(), BridgeBuilder.class );
+		}
+		else {
+			// The bridge will be auto-detected from the property type
+			return null;
```
gsmet: And same remark here about using `Optional`.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/PropertyModel.java
```diff
@@ -6,12 +6,19 @@
...
+
 public interface PropertyModel<T> {
 
 	String getName();
 
 	Class<T> getJavaType();
 
+	<A extends Annotation> Stream<A> getAnnotationsByType(Class<A> annotationType);
+
+	Stream<? extends Annotation> getAnnotationsByMetaAnnotationType(Class<? extends Annotation> metaAnnotationType);
```
gsmet: I'm not sure about it but I have a vague recollection of `Stream#toString()` being unreadable. If so, I must admit that I'm not very fond of using `Stream` as return types.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/TypeModel.java
```diff
@@ -6,10 +6,22 @@
...
 
+	<A extends Annotation> Optional<A> getAnnotationByType(Class<A> annotationType);
+
+	<A extends Annotation> Stream<A> getAnnotationsByType(Class<A> annotationType);
+
+	Stream<? extends Annotation> getAnnotationsByMetaAnnotationType(Class<? extends Annotation> metaAnnotationType);
+
 	PropertyModel<?> getProperty(String propertyName);
 
+	Stream<PropertyModel<?>> getDeclaredProperties();
```
gsmet: Same remark about using `Stream` as return types.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/util/spi/AnnotationHelper.java
```diff
@@ -0,0 +1,105 @@
...
+			AnnotatedElement annotatedElement, Class<A> annotationType) {
+		return Optional.ofNullable( annotatedElement.getAnnotation( annotationType ) );
+	}
+
+	public <A extends Annotation> Stream<A> getAnnotationsByType(
+			AnnotatedElement annotatedElement, Class<A> annotationType) {
+		return Arrays.stream( annotatedElement.getAnnotationsByType( annotationType ) );
+	}
+
+	public Stream<? extends Annotation> getAnnotationsByMetaAnnotationType(
```
gsmet: Same remark about using Stream as return type.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
yrodiere: > As null seems to be a valid use case we need to take care of.

Yes it is.

But... this is a private method, and its return type will be used directly as an argument to `PojoPropertyNodeMappingCollector#identifierBridge`, which is also internal code.

If I did return an `Optional`, I would have to change `PojoPropertyNodeMappingCollector#identifierBridge` to accept an `Optional`. And then comes the controversy of "do not use `Optional`s as method parameters": https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments
(By the way, IDEA displays a warning if you try to do that)

All that thinking seems a bit much just for internal code. Frankly I don't know where to stand here. Don't even know if it's worth discussing, especially for internal code. Maybe we should decide whether or not we want `Optional` parameters once for all? Some policy like "unless there's a good reason (performance-critical code, API readability), use `Optional` wherever a parameter is optional" or "never use `Optional` in method parameters".

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
yrodiere: I thought the name was already long enough, and "bridge" would be redundant because the whole method is about building an identifier bridge builder. I mean, if we go down this path I could also name it `identifierBridgeBuilder...`...
But as you wish.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/PropertyModel.java
yrodiere: Thing is, it comes really handy when we need to manipulate the "set" of elements, and here we do... a lot. See `org.hibernate.search.v6poc.entity.orm.model.impl.HibernateOrmPropertyModel#getAnnotationsByMetaAnnotationType` (aggregate multiple element sets), `org.hibernate.search.v6poc.entity.orm.model.impl.HibernateOrmIntrospector#getAnnotationsByMetaAnnotationType` (expand some elements to multiple elements, filter the set of elements)...
I guess we could do the filtering "in place", using a `List` or `Set`, but I'm not sure this would be particularly easier to read.

As to debugging... You can just evaluate `<the call that returns a new Stream>.toArray()` and you're good to go. Of course you have to make sure to call to array on a new `Stream` instance, not the one being used in the code, but in this case it doesn't seem to be a problem... ?

engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
```diff
@@ -93,7 +93,7 @@ public SearchMappingRepository build() {
 		// TODO close the holder (which will close the backends) if anything fails after this
 
 		TypeMetadataCollectorImpl metadataCollector = new TypeMetadataCollectorImpl();
-		contributors.forEach( c -> c.contribute( metadataCollector ) );
+		contributors.forEach( c -> c.contribute( buildContext, metadataCollector ) );
```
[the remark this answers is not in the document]
gsmet: I don't know the requirements of my future self. I just know that having a build broken at a certain point in time is pretty bad and can be annoying for my future self so I usually try to avoid that when I can.

mapper-javabean/src/main/java/org/hibernate/search/v6poc/entity/javabean/model/impl/JavaBeanTypeModel.java
gsmet: Ah ah sorry, I missed it was a standard Java class. Never saw this one before.

mapper-javabean/src/main/java/org/hibernate/search/v6poc/entity/javabean/model/impl/JavaBeanTypeModel.java
gsmet: Yeah, it was not about sharing code but more than it has been a common requirement these days so making it easy for the user to have another strategy (for common cases) might be interesting.

It was more a thought I wanted to share than anything else.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/PojoMappingContributor.java
gsmet: Mmmh, maybe I missed something but when I reviewed, I just saw this import being removed, without any code removal. I reviewed commit per commit so maybe it's because of that.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
gsmet: I would have .orElseThrow( just below the above (. Anyway, it's no big deal.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationPojoTypeNodeMetadataContributorImpl.java
gsmet: OK. I must admit that I don't make that much difference between internal code or not. I just try to have all the code following the same standard and be readable.

I'm a bit worried that if we sometimes use `Optional` and sometimes not, we will tend to forget about taking care of the `null` case. But I probably worry too much.

Let's do it this way for now.

yrodiere:
Merged, thanks! I added a section about the Stream/Set topic in the POC document, so that we will think of revisiting this later.

PR #8, opened 2018-02-05
 Add sorts to the query DSL

yrodiere:
I need this to rewrite the tests, since validating results without a pre-defined order would be challenging :)
So here comes a simple API for sorts, and a similarly simple implementation for Elasticsearch.

I say "simple" because I didn't try to address the numerous types of sort we could envision, not least the spatial distance sort and the [nested sort](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html#nested-sorting). That will have to wait, as I have more pressing concerns (i.e. writing a showcase to explain how this all will look like in a real-life application).

Sort examples can be found in `JavaBeanElasticsearchAnnotationMappingIT#search` in particular. It works similarly to the search predicates:

```
SearchQuery<PojoReference> query = manager.search(
				Arrays.asList( IndexedEntity.class, YetAnotherIndexedEntity.class )
		)
		.query()
		.asReferences()
		.predicate(
				// ...
		)
		.sort( c ->
				c.byField( "numeric" ).desc()
						.then().byField( "myLocalDateField" )
						.then().byScore()
		)
```


gsmet:
Looks good to me at a quick glance. Not totally convinced by the use of lambdas but we can discuss that later. Feel free to merge it!

yrodiere:
Merged, thanks

PR #10, opened 2018-02-12
Add test coverage reports using JaCoCo and Coveralls

yrodiere:
Reports should appear there, hopefully a few minutes after I submitted this PR: https://coveralls.io/github/hibernate/hibernate-search-6-poc

Example with my own fork: https://coveralls.io/github/yrodiere/hibernate-search-6-poc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -312,6 +312,7 @@
                         <reuseForks>true</reuseForks>
                         <redirectTestOutputToFile>true</redirectTestOutputToFile>
                         <runOrder>alphabetical</runOrder>
+                        <argLine>${surefireCoverageArgLine}</argLine>
                     </configuration>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -303,25 +327,98 @@
...
+                        <redirectTestOutputToFile>true</redirectTestOutputToFile>
+                        <runOrder>alphabetical</runOrder>
+                        <argLine>${surefireLateEvaluationArgLine}</argLine>
+                    </configuration>
+                </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-failsafe-plugin</artifactId>
+                    <!-- 2.19.1 seems buggy, see https://hibernate.atlassian.net/browse/HV-1117 -->
+                    <version>2.18.1</version>
```
gsmet: I think it's the same issue we had with Surefire so we can probably upgrade it too?

reports/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+            <artifactId>hibernate-search-integration-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
```
gsmet: TBH, I would prefer if we didn't execute them by default.

yrodiere:
Addressed your comments and merged. Thanks!

PR #11, opened 2018-02-14
Test overhaul and showcase

yrodiere:
Main changes:

* Use an actual client (instead of a stubbed one) in the Elasticsearch backend, allowing to perform integration tests with an actual Elasticsearch cluster
* Introduce very limited support for analyzer/normalizers (necessary for tests)
* Introduce a way to declare fields as sortable (necessary for tests)
* Move backend tests to a mapper-independent TCK. This TCK can be used with other backends relatively easily (see the src/test/resources and pom.xml of the Elasticsearch integration tests)
* Make mapper tests backend-independent

The showcase is not really complete yet, but I will need to add a few features before it is complete. So let's merge it for now, and I'll change it as I introduce the new features.


yrodiere:
Merged, thanks!

PR #12, opened 2018-02-15
Remove the need for ConcurrentReferenceHashMap for storing pojo workers for in-progress transactions

yrodiere:
This class was huge and had no unit test. Besides, Sanne wanted to get
rid of it, so let's try something.

The fact that the map is now stored as a transient property inside the
session shouldn't lead to any *new* problem, since we used to rely on a
map with weak keys, and the transaction identifier inside the session is
transient. So previously, when a Session was serialized, the transaction
identifier was only referenced from the map, and thus the value
(containing all the works we want to process on commit) was lost.

We probably cannot do that in Search 5... at least it would be more complex. I think it would require changing the `TransactionContext` API. For future reference though, the classes to look at would be `TransactionContext`, `PerTransactionWorker` and `EventSourceTransactionContext`.


gsmet:
Looks like a sane approach. Rebased and merged!

PR #14, opened 2018-02-19
Fix a few backend field bugs

yrodiere:
Pushing that now just so you don't waste time when comparing the behavior of your Lucene backend with the stub or the Elasticsearch backend, @gsmet .

I will add tests as part of a PR that I will submit in a few days (it's not ready yet, but for reasons unrelated to these fixes).


yrodiere:
... And pushed an update to fix checkstyle errors.

gsmet:
Merged thanks!

PR #17, opened 2018-02-20
Rename all() predicate to matchAll()

gsmet:


engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/impl/MatchAllPredicateContextImpl.java
```diff
@@ -12,50 +12,50 @@
...
 			Supplier<N> nextContextProvider) {
 		this.factory = factory;
 		this.nextContextProvider = nextContextProvider;
-		this.builder = factory.all();
+		this.builder = factory.matchAll();
 	}
 
 	@Override
-	public AllPredicateContext<N> boostedTo(float boost) {
+	public MatchAllPredicateContext<N> boostedTo(float boost) {
```
gsmet: @yrodiere btw, does it make sense to have a boostedTo() for matchAll()?

yrodiere: In Elasticsearch it defines the score of that query. Not sure if it's useful, but it seems to make sense, yes.
See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-all-query.html

engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/SearchPredicateContainerContext.java
```diff
@@ -20,7 +20,7 @@
  */
 public interface SearchPredicateContainerContext<N> {
 
-	AllPredicateContext<N> all();
+	MatchAllPredicateContext<N> all();
```
yrodiere: Maybe we should rename this to `matchAll()` too... But I guess it's a different story, since it's user API. I'll merge anyway.

yrodiere:
Merged, thanks!

PR #19, opened 2018-02-20
Add basic (non-configurable) support for container value extractors in @IndexedEmbedded and @Field

yrodiere:
This is a first step toward introducing actual support for container value extractors.

Current limitations (to be addressed in another PR):

 * value extractors are selected automatically based on the property type, and cannot be overridden
 * value extractors cannot be used in custom bridges (implementations of the `Bridge` interface)
 * value extractors probably don't work on generic types extending Iterable/Collection/Map, but changing the position of the generic type parameter, nor on types extending a fully-parameterized generic interface (`MyCollection extends Collection<String>` for instance). This is because the generic type reflection is, at the moment, very naïve.


gsmet:
Merged, thanks!

PR #20, opened 2018-02-21
Split Bridge into two interfaces: TypeBridge and PropertyBridge

yrodiere:
As discussed during the face-to-face. 

*WARNING*: this PR is based on #19 , which should be merged first.

This is a first step, I'm not really happy about the duplication of the javadoc in TypeBridge/PropertyBridge, which gives away the fact that those bridges have multiple responsibilities. I will try to improve on this.

But still, it's better than it was, so... let's merge?


gsmet:
Merged, too, thanks.

PR #21, opened 2018-02-21
Close processors and trim down the processor tree as necessary

yrodiere:
Submitting the two changes in a single PR because it's hard to test one without the other.

The first commit is trivial: just close the resources like I should have from the beginning.

The second one is a bit less obvious, but hopefully the commit comment should explain it.


gsmet:
It looks like a nice improvement. Merged, thanks.

PR #22, opened 2018-02-23
TCK - Prefix the index name with the class and test names

gsmet:
@yrodiere here is what we discussed this evening.


yrodiere:
Merged, thanks!

PR #24, opened 2018-02-28
Tests and fixes related to Hibernate ORM access type and annoted type discovery

yrodiere:


engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
```diff
@@ -142,34 +146,62 @@ public SearchMappingRepository getBuiltResult() {
...
 
 		MapperContribution(MapperFactory<C, M> mapperFactory) {
 			this.mapperFactory = mapperFactory;
 		}
 
 		public void update(MappableTypeModel typeModel, String indexName, C contributor) {
+			if ( freezedTypes.contains( typeModel ) ) {
+				throw new AssertionFailure(
+						"Attempt to add a mapping contribution for a type which has already been mapped."
+						+ " There is a bug in the mapper implementation."
```
gsmet: Might be a good idea to add some info about the typeModel to the exception.

engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
```diff
@@ -142,34 +146,62 @@ public SearchMappingRepository getBuiltResult() {
...
+			 * which may discover new embedded types while mapping a type.
+			 *
+			 * We do not, however, allow new contributions to types already encountered in this method
+			 * (the "freezed" types).
+			 */
+			while ( !typesLeftToProcess.isEmpty() ) {
+				typesBeingProcessed.clear();
+				typesBeingProcessed.addAll( typesLeftToProcess );
+				typesLeftToProcess.clear();
+				freezedTypes.addAll( typesBeingProcessed );
```
gsmet: I don't think it has importance but aren't the types freezed after the execution of the following code.

I don't think it changes anything but it would seem more logical to me.

integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/OrmAnnotationMappingAccessTypeIT.java
```diff
@@ -36,6 +37,7 @@
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import junit.framework.AssertionFailedError;
```
gsmet: Maybe it's just in HV but we usually avoid the class from the `junit.`, wondering if `AssumptionViolationException` would work for you?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationMappingDefinitionImpl.java
```diff
@@ -49,21 +51,56 @@ public AnnotationMappingDefinition add(Set<Class<?>> annotatedTypes) {
...
-					collector.collect( mapperFactory, typeModel, indexName,
-							new AnnotationPojoTypeNodeMetadataContributorImpl( beanResolver, typeModel ) );
-				} );
+				.forEach( contributor::contributeTypeAndSuperTypes );
 	}
 
+	/**
+	 * A metadata contributor that will call itself recursively on newly discovered types.
+	 */
+	private static class DiscoveringContributor implements PojoMappedTypeListener {
```
gsmet: The name of this class is very vague, even for an internal class.

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/AbstractHibernateOrmTypeModel.java
```diff
@@ -114,9 +114,15 @@ public boolean isSubTypeOf(MappableTypeModel other) {
 						return getProperty( propertyName );
 					}
 					catch (PropertyNotFoundException e) {
+						// Error resolving the property through Hibernate internals
 						// Ignore this property
 						return (PojoPropertyModel<?>) null;
```
gsmet: Very minor but you have a cast here and not in the catch just below.

integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/OrmAnnotationMappingAccessTypeIT.java
yrodiere: Right, I'll change that to `Assert.fail`. No, `AssumptionViolationException` would not do, it's meant to ignore tests conditionally (e.g. for specific DB dialects), not to fail tests.

gsmet:
@yrodiere Travis is failing with this error:
```
[ERROR] /home/travis/build/hibernate/hibernate-search-6-poc/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/building/impl/PojoTypeNodeMappingCollector.java:12:8: Unused import - org.hibernate.search.v6poc.entity.pojo.model.spi.PojoTypeModel. [UnusedImports]
```

Shouldn't be too hard to fix :).

engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
yrodiere: They are actually freezed before, because as soon as we started processing them, we don't want any more contribution to these types. The "freezing" here is about rejecting new contributions, not about marking the contributions as processed.

engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
yrodiere: Right, done.

engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
yrodiere: I will make it more explicit in the error message.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/AnnotationMappingDefinitionImpl.java
yrodiere: Changed to `LazilyDiscoveringMetadataContributor`. If it's still too vague, well... I don't know :/

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/AbstractHibernateOrmTypeModel.java
yrodiere: Yeah, it the cast on the first return was required by the compiler, which doesn't know what the return type of the lambda is without it. If I put a cast on the second return, I get a warning telling me it's not necessary...

yrodiere:
@gsmet Thanks! I updated the PR. I felt bad about the lazy annotation mapping discovery feature being unclear, so I added dedicated tests (commit "Test lazy discovery of annotated types"). Of course I found bugs... So I changed the implementation of lazy annotation mapping discovery a bit, and I tried to tidy it up a bit while I was at it. Hopefully it's clearer now...

yrodiere:
Merged as per our discussion on Hangouts. Thanks!

PR #26, opened 2018-03-16
Make annotation discovery a global setting

yrodiere:
The previous implementation tried to be smart, discovering nested type
annotations when processing annotation mapping, and ignoring them when
processing programmatic mapping.

But the approach was fundamentally flawed: some of the discovered
metadata is cached, and shared between all the mappings. For instance
this is the case of markers. So with the previous approach, discovered
annotations could impact a programmatic mapping or not depending on
whether they were discovered before or after the programmatic mapping
was processed. Which can be terribly confusing.

Thus I gave up and made annotation discovery a global setting: either
you enable it everywhere (in annotation and programmatic mappings),
or you don't enable it at all.

The original goal of this switch was to allow people with annotated
classes to ignore the annotations and only rely on a programmatic
mapping, so I also added a setting in the ORM integration to disable
annotation mapping completely.


yrodiere:
Discussed with Guillaume over HipChat and merged.

PR #27, opened 2018-03-19
Rename FunctionBridge to ValueBridge

yrodiere:
The new name is more consistent with TypeBridge and PropertyBridge:
where TypeBridge works on a type and its metadata, and PropertyBridge
works on a property and its metadata, ValueBridge does not have access
to any metadata, only the value, and it works on that value.


gsmet:
Do you really expect me to review that? :)

I think you can merge it.

yrodiere:
Fair enough :) Merged!

PR #29, opened 2018-03-26
First implementation of the Lucene backend

gsmet:
A few things are still very naive compared to what we support in Search 5 though.


gsmet:
@yrodiere I'm a bit confused about the coverall reports. It's saying that none of the added lines are tested whereas the integration tests should at least exercise some of them...

yrodiere:
@gsmet You need to add the Lucene backend integration tests to the coverage report. I added a commit, it should be better now.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/search/query/impl/SearchQueryFactoryImpl.java
```diff
@@ -0,0 +1,139 @@
...
+			}
+		}
+		return new ProjectionHitExtractor( documentExtractors );
+	}
+
+	private <C, T> SearchQueryBuilderImpl<C, T> createSearchQueryBuilder(
+			SessionContext sessionContext, HitExtractor<? super C> hitExtractor, HitAggregator<C, List<T>> hitAggregator) {
+
+
+		return new SearchQueryBuilderImpl<>(
```
yrodiere: The two blank lines above seem unnecessary.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/search/query/impl/SearchResultExtractorImpl.java
```diff
@@ -0,0 +1,53 @@
...
+		for ( ScoreDoc hit : topDocs.scoreDocs ) {
+			C hitCollector = hitAggregator.nextCollector();
+			hitExtractor.extract( hitCollector, indexSearcher, hit );
+		}
+
+		final List<T> finalHits = Collections.unmodifiableList( hitAggregator.build() );
+		return new SearchResult<T>() {
+			@Override
+			public long getHitCount() {
+				return topDocs.totalHits;
```
yrodiere: You might want to copy this value to a local variable first, so that your anonymous class doesn't keep a reference to the `TopDocs`.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/search/reader/impl/MultiReaderFactory.java
```diff
@@ -0,0 +1,62 @@
...
+
+	private MultiReaderFactory() {
+		//not allowed
+	}
+
+	public static IndexReader openReader(Set<String> indexNames, Set<ReaderProvider> readerProviders) {
+		if ( readerProviders.size() == 0 ) {
+			return null;
+		}
+		else {
```
yrodiere: I know it's pre-existing code, but maybe we should add a special case for when there's only one ReaderProvider?

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/work/impl/AddEntryLuceneWork.java
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	@Override
+	public CompletableFuture<Long> execute(LuceneWorkExecutionContext context) {
+		return Futures.create( () -> addEntry( context.getIndexWriter() ) );
+	}
+
+	private CompletableFuture<Long> addEntry(IndexWriter indexWriter) {
+		try {
+			return CompletableFuture.completedFuture( indexWriter.addDocuments( indexEntry ) );
```
yrodiere: The fact that this method (and similar methods in other works) is blocking should really be mentioned with a big "FIXME"  comment...

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/work/impl/ExecuteQueryLuceneWork.java
```diff
@@ -0,0 +1,57 @@
...
+
+	private final LuceneSearcher<T> searcher;
+
+	public ExecuteQueryLuceneWork(LuceneSearcher<T> searcher) {
+		this.searcher = searcher;
+	}
+
+	@Override
+	public CompletableFuture<SearchResult<T>> execute(LuceneQueryWorkExecutionContext context) {
+		return Futures.create( () -> CompletableFuture.completedFuture( executeQuery( searcher ) ) );
```
yrodiere: Same as Index works, the fact that this method is blocking should really be mentioned in a "FIXME" comment.

integrationtest/backend-tck/src/main/java/org/hibernate/search/v6poc/integrationtest/backend/tck/util/TckConfiguration.java
```diff
@@ -42,11 +46,25 @@ private TckConfiguration() {
 		catch (IOException e) {
 			throw new IllegalStateException( "Error loading TCK properties file: " + PROPERTIES_PATH );
 		}
+
+		addTimestampToLuceneRootDirectory( properties );
```
yrodiere: I really think we should instead replace occurrences of the `#{tck.startup.timestamp}` string with the actual timestamp in every property. I've tried hard to not tie the TCK to a particular backend implementation and I'd really like it to stay that way...
Note that while we're at it, we could also replace occurrences of `#{tck.test.id}` and instantiate the TckConfiguration once per test. It would allow us to end the practice of using the test ID in index names, because we could just inject the test ID in the Lucene base directory path.

yrodiere:
You might want to have a look at the coverage report, it gives some interesting insights.

What I've found so far:

* `org.hibernate.search.v6poc.backend.lucene.document.model.impl.LuceneObjectFieldModel` is never used
* There isn't any test for `org.hibernate.search.v6poc.backend.lucene.search.predicate.impl.LuceneSearchPredicateFactory#fromLuceneQuery`
* There isn't any test for `org.hibernate.search.v6poc.backend.lucene.search.sort.impl.LuceneSearchSortFactory#fromLuceneSortField`

... And unrelated to your PR,  we should seriously think about adding tests for multi-index, for multi-tenancy, for stream works, for document updates and for document deletions.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/impl/LuceneBackendFactory.java
```diff
@@ -32,9 +32,8 @@
 			.asString()
 			.build();
 
-	// TODO GSM: probably better to introduce a parser for Paths
-	private static final ConfigurationProperty<Optional<String>> ROOT_DIRECTORY = ConfigurationProperty.forKey( "lucene.root_directory" )
-			.asString()
+	private static final ConfigurationProperty<Optional<Path>> ROOT_DIRECTORY = ConfigurationProperty.forKey( "lucene.root_directory" )
+			.as(Path.class, Paths::get)
```
yrodiere: You're missing some spaces around parentheses.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/search/predicate/impl/MatchPredicateBuilderImpl.java
```diff
@@ -0,0 +1,46 @@
...
+	private Object value;
+
+	public MatchPredicateBuilderImpl(String absoluteFieldPath, LuceneFieldQueryBuilder queryBuilder) {
+		this.absoluteFieldPath = absoluteFieldPath;
+		this.queryBuilder = queryBuilder;
+	}
+
+	@Override
+	public void value(Object value) {
+		this.value = value;
```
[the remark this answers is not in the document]
yrodiere: Yes and no, the predicate implementations could still be in the same package, and wouldn't need to have a dependency to the model. They would just be instantiated by the "FieldQueryBuilderFactory".

If you are talking about the dependency from `SearchPredicateFactoryImpl` to the model, it's already there.

Ok, let's discuss this later, I'll add it to the google doc.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/logging/impl/Log.java
```diff
@@ -0,0 +1,135 @@
...
+	@Message(id = 5, value = "Root directory '%2$s' of backend '%1$s' exists but is not a writable directory.")
+	SearchException localDirectoryBackendRootDirectoryNotWritableDirectory(String backendName, Path rootDirectory);
+
+	@Message(id = 6, value = "Unable to create root directory '%2$s' for backend '%1$s'.")
+	SearchException unableToCreateRootDirectoryForLocalDirectoryBackend(String backendName, Path rootDirectory, @Cause Exception e);
+
+	@Message(id = 7, value = "Undefined Lucene directory provider for backend '%1$s'.")
+	SearchException undefinedLuceneDirectoryProvider(String backendName);
+
+	@Message(id = 8, value = "Unrecognized Lucene backend type '%2$s' for backend '%1$s'.")
```
yrodiere: "backend type" => "directory provider", you missed this one.

gsmet:
Merged, thanks for the thorough review.

I will follow up on the various remaining issues (and add some tests for the untested features).

PR #31, opened 2018-03-26
Use jqassistant to avoid implementation leakages

gsmet:
@yrodiere as discussed the other day, I implemented some JQAssistant checks.

We have a few failures. Maybe we have to tighten up the rules but they look legit to me.

```
[ERROR] --[ Constraint Violation ]-----------------------------------------
[ERROR] Constraint: my-rules:PublicMethodsMayNotExposeInternalTypes
[ERROR] Severity: MAJOR
[ERROR] API/SPI methods must not expose internal types.
[ERROR]   method=org.hibernate.search.v6poc.entity.pojo.mapping.building.spi.PojoMapperFactory#org.hibernate.search.v6poc.entity.pojo.mapping.building.impl.PojoMapper createMapper(org.hibernate.search.v6poc.engine.spi.BuildContext,org.hibernate.search.v6poc.cfg.ConfigurationPropertySource)
[ERROR]   method=org.hibernate.search.v6poc.entity.javabean.JavaBeanMappingContributor#org.hibernate.search.v6poc.entity.javabean.JavaBeanMapping toReturnType(org.hibernate.search.v6poc.entity.javabean.mapping.impl.JavaBeanMappingImpl)
[ERROR]   method=org.hibernate.search.v6poc.entity.orm.mapping.HibernateOrmMappingContributor#org.hibernate.search.v6poc.entity.orm.mapping.HibernateOrmMapping toReturnType(org.hibernate.search.v6poc.entity.orm.mapping.impl.HibernateOrmMappingImpl)
[ERROR] -------------------------------------------------------------------

[ERROR] --[ Constraint Violation ]-----------------------------------------
[ERROR] Constraint: my-rules:PublicTypesMayNotExtendInternalTypes
[ERROR] Severity: MAJOR
[ERROR] API/SPI types must not extend/implement internal types.
[ERROR]   class=org.hibernate.search.v6poc.search.dsl.predicate.spi.SearchTargetPredicateRootContext
[ERROR]   class=org.hibernate.search.v6poc.search.dsl.sort.spi.SearchTargetSortRootContext
[ERROR] -------------------------------------------------------------------
```


yrodiere:
@gsmet I added a few commits to fix the errors, let's see if Travis agrees with me

yrodiere:
It worked :) I'll wait for your input on my changes, but to me it looks good to merge.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/building/spi/PojoMapperFactory.java
```diff
@@ -31,7 +32,8 @@ protected PojoMapperFactory(PojoBootstrapIntrospector introspector, boolean impl
 	}
 
 	@Override
-	public final PojoMapper<M> createMapper(BuildContext buildContext, ConfigurationPropertySource propertySource) {
+	public final Mapper<PojoTypeMetadataContributor, M> createMapper(BuildContext buildContext,
```
gsmet: Considering `PojoTypeMetadataContributor` is an impl too, I don't think it solves the issue.

engine/src/main/java/org/hibernate/search/v6poc/engine/SearchMappingRepository.java
```diff
@@ -10,9 +10,6 @@
 import org.hibernate.search.v6poc.engine.impl.SearchMappingRepositoryBuilderImpl;
 import org.hibernate.search.v6poc.entity.mapping.spi.MappingKey;
 
-/**
- * @author Yoann Rodiere
- */
```
gsmet: I don't understand why you remove these tags now that they are everywhere. It's really not helping when reviewing.

yrodiere: I'm only removing them when I made useful changes to the same file.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/building/spi/PojoMapperFactory.java
yrodiere: Right, this particular change was useless, I removed it. And there was a reference to it in a useless protected method in `PojoMappingContributorImpl`, I removed it too.
But it doesn't matter: the commit moves `PojoMapperFactory` to impl and removes the need for SPIs to refer to the `MapperFactory`, so it does solve the problem.

yrodiere: I just rebased and pushed.

engine/src/main/java/org/hibernate/search/v6poc/engine/SearchMappingRepository.java
yrodiere: I guess this one slipped in because I reverted some changes... Sorry about that.

gsmet:
Checkstyle is unhappy:
```
[ERROR] /home/travis/build/hibernate/hibernate-search-6-poc/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/building/impl/PojoMapperFactory.java:11:8: Unused import - org.hibernate.search.v6poc.entity.mapping.building.spi.Mapper. [UnusedImports]
```

engine/src/main/java/org/hibernate/search/v6poc/engine/SearchMappingRepository.java
gsmet: If you think they should go away, let's remove them once and for all in a single commit. But we should do it after the Lucene merge because I added the tags there too.

Personally, I don't have an issue with them: we maintain the `@author` tags for HV and I don't think it's that useless.

yrodiere: > If you think they should go away, let's remove them once and for all in a single commit. But we should do it after the Lucene merge because I added the tags there too.

Ok, I'll do that.

yrodiere:
Fixed, sorry. Is it good apart from that? Can I merge?

engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryImpl.java
```diff
@@ -15,24 +15,20 @@
...
+	SearchMappingRepositoryImpl(Map<MappingKey<?>, MappingImplementor<?>> mappings, Map<String, Backend<?>> backends) {
 		this.mappings = mappings;
 		this.backends = backends;
 	}
 
 	@Override
 	public <M> M getMapping(MappingKey<M> mappingKey) {
 		@SuppressWarnings("unchecked") // See SearchMappingRepositoryBuilderImpl: we are sure that, if there is a mapping, it implements M
-		M mapping = (M) mappings.get( mappingKey );
+		M mapping = (M) mappings.get( mappingKey ).toAPI();
```
gsmet: I don't see the purpose of this `toAPI()` method? I mean your objects are inheriting from the API type so why would you need this method?

yrodiere: I don't want to expose the MappingImplementor interface through the APIs, but I need the mappings to implement it for the benefit of the SearchMappingRepository.
In particular I don't want users to start closing the mappings, because it's the responsibility of the SearchMappingRepository.

yrodiere: To clarify: the type M here will be a user-facing interface, not the implementation type. So the user-facing will not extend MappingImplementor, but the implementation type will.

gsmet:
Merged!

PR #32, opened 2018-03-29
Only use one global StoredFieldVisitor for field extraction

gsmet:


yrodiere:
> Lines: +156 −298

That's the kind of diff I like :)

Merged, thanks!

PR #33, opened 2018-03-29
Move utils to dedicated modules in a "util" directory

yrodiere:
@gsmet No, I don't expect you to review the diff :) But maybe you could check out this branch and have a look at the resulting module structure, see if it makes sense?


gsmet:
Hum. I think I would have put the integration tests utils and shared resources into the integrationtest/ directory.

That makes more sense because you can completely ignore them if you're not interested in the integration tests.

yrodiere:
On the other hand you can ignore them too with my proposal, since they are in a separate subdirectory.

On the plus side:

* the integration test utils are no longer lost between integration test modules, so they will not be mistaken for "just some other integration tests".
* this structure has the advantage of allowing to add integration tests for utils in the `integrationtest` directory, should we ever need any. (Theoretically we should add some for the ORM-related utils, but you know...)
* this structure will be more consistent once we add the utils for unit tests (there are some in Search 5, so I guess there will be some in Search 6 as well).
* this structure will be more consistent and easier to work with if we decide to publish IT-related utils for use by Hibernate Search users, especially since we will probably use those in our own tests: the public and internal util modules will be closer to each other.

So I'm not sure moving the utils back to the `integrationtest` directory is such a good idea...

gsmet:
Let's agree to disagree. I think having the integration tests as a single thing is better in the long run.

Anyway, let's merge it as it will soon be in the way.

yrodiere:
Merged, thanks.

PR #34, opened 2018-03-30
Make sorting contribution of types more flexible

gsmet:
As discussed in the previous PR.

### DIFF
--- a/backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/impl/IndexSchemaFieldGeoPointContext.java
+++ b/backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/impl/IndexSchemaFieldGeoPointContext.java
@@ -16,7 +16,6 @@ import org.apache.lucene.document.Document;
 import org.apache.lucene.document.LatLonPoint;
 import org.apache.lucene.document.StoredField;
 import org.apache.lucene.index.IndexableField;
-import org.apache.lucene.search.SortField.Type;
 import org.hibernate.search.v6poc.backend.document.impl.DeferredInitializationIndexFieldAccessor;
 import org.hibernate.search.v6poc.backend.document.model.Store;
 import org.hibernate.search.v6poc.backend.lucene.document.impl.LuceneDocumentBuilder;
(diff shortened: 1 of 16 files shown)


backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/impl/LuceneIndexSchemaFieldNode.java
```diff
@@ -50,6 +55,18 @@ public LuceneFieldQueryFactory getQueryFactory() {
...
+		if ( !Objects.equals( sortContributor, other.sortContributor ) || !Objects.equals( formatter, other.formatter )
+				|| !Objects.equals( queryFactory, other.queryFactory ) ) {
+			return false;
+		}
+		return true;
+	}
+
 	@Override
 	public String toString() {
 		return getClass().getSimpleName() + "[parent=" + parent + ", fieldName=" + fieldName + "]";
```
yrodiere: Since you are using this `toString()` implementation in logs to explain formatting conflicts, it should probably include some information about the format.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/impl/IndexSchemaFieldLocalDateContext.java
```diff
@@ -169,16 +156,19 @@ public int hashCode() {
...
 
 	private static final class LocalDateFieldQueryFactory implements LuceneFieldQueryFactory {
 
 		private final LocalDateFieldFormatter localDateFieldFormatter;
 
+		private final int hashCode;
+
 		private LocalDateFieldQueryFactory(LocalDateFieldFormatter localDateFieldFormatter) {
 			this.localDateFieldFormatter = localDateFieldFormatter;
+			this.hashCode = buildHashCode();
```
yrodiere: From what I can see we don't even use the hashcode for this type (even indirectly through `LuceneIndexSchemaFieldNode`), so this looks like very premature optimisation :) But as you wish.

gsmet: Yeah, it's an habit coming from HV but let's keep it for the elements appearing in a Map. Removed.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/impl/IndexSchemaFieldGeoPointContext.java
```diff
@@ -41,7 +40,8 @@ protected void contribute(DeferredInitializationIndexFieldAccessor<GeoPoint> acc
 				parentNode,
 				getFieldName(),
 				new LuceneGeoPointFieldFormatter( parentNode.getAbsolutePath( getFieldName() ), getStore() ),
-				null // for now we don't have a query builder for GeoPoint
+				null, // for now we don't have a query builder for GeoPoint
```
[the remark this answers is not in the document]
gsmet: Sure, fixed.

PR #35, opened 2018-03-30
Reorganize loggers

gsmet:
As discussed on HipChat.

On top of https://github.com/hibernate/hibernate-search-6-poc/pull/34.


gsmet:
Apparently, download.oracle.com is not accessible so the Travis builds fail...

yrodiere:
I think you missed at least one pre-existing message; this is from Search 5:

```
	@LogMessage(level = WARN)
	@Message(id = 55, value = "Unable to close Lucene IndexReader")
	void unableToCloseLuceneIndexReader(@Cause Exception e);
```

And we have:

```
 	@LogMessage(level = Level.WARN)
 	@Message(id = 523, value = "Unable to close the index reader for index '%2$s' of backend '%1$s'.")
 	void unableToCloseIndexReader(String backendName, String indexName, @Cause Exception e);
```

I didn't check everything, there may be others.

But maybe the goal was not to find them all, and you wanted to do that later?

gsmet:
Fixed 2 preexisting messages and rebased.

@yrodiere I think it should be good now.

yrodiere:
Merged, thanks!

PR #37, opened 2018-04-06
Cleanup of code related to casters and subtype testing

yrodiere:
This is useful for the upcoming PR introducing automatic reindexing, but makes sense even without that PR.


gsmet:
Merged, thanks!

PR #41, opened 2018-04-19
HSEARCH-3128 Search 6 groundwork - Allow per-backend access to the Elasticsearch RestClient

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3128


integrationtest/backend-elasticsearch/src/test/java/org/hibernate/search/v6poc/integrationtest/backend/elasticsearch/ExtensionIT.java
```diff
@@ -263,6 +279,30 @@ public void sort_fromJsonString_separateSort() {
...
+
+		// Test that the client actually works
+		Response response = restClient.performRequest( "GET", "/" );
+		Assertions.assertThat( response.getStatusLine().getStatusCode() ).isEqualTo( 200 );
+	}
+
+	@Test
+	public void backend_getClient_error_invalidClass() {
+		Backend backend = mappingRepository.getBackend( BACKEND_NAME );
+		ElasticsearchBackend elasticsearchBackend = backend.withExtension( ElasticsearchExtension.get() );
```
gsmet: I saw your discussion with Sanne. In this particular case, I wouldn't use the extension framework and use a simple `as()` as discussed with Sanne.

yrodiere: I guess I'm outnumbered. Well, we'll always be able to reintroduce it later if we find a compelling reason other than consistency.
Added a commit to remove the backend extensions. I named the method `unwrap` though, in order to at least be consistent with JPA's `EntityManager` and Search 5's `SearchFactory`.

gsmet:
Merged, thanks!

PR #43, opened 2018-04-24
HSEARCH-3123 Search 6 groundwork - Clean up the backend internals

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3123

This removes several cyclic dependencies in backends, while still not requiring to pass lots of objects to constructors, which was the main argument in favor of exposing getters in the Backend types, IIRC.

This also splits some packages and make their purpose clearer, in my opinion. See "HSEARCH-3123 Move model DSL types to their own packages" in particular.

The reason for each change is explained in each commit's message.

Note that this does not solve everything, but the main problems I spotted. Others are relatively minor, or require a lot more work and will be affected by future works anyway.


backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/index/impl/ElasticsearchIndexSearchTargetBuilder.java
```diff
@@ -50,14 +50,14 @@ public IndexSearchTarget build() {
...
-		return new ElasticsearchIndexSearchTarget( backend.getSearchContext(), searchTargetModel );
+		return new ElasticsearchIndexSearchTarget( searchBackendContext, searchTargetModel );
 	}
 
 	@Override
 	public String toString() {
 		return new StringBuilder( getClass().getSimpleName() )
 				.append( "[" )
-				.append( "backend=" ).append( backend )
+				.append( "searchBackendContext=" ).append( searchBackendContext )
```
gsmet: AFAICS it does not have a proper `toString()`, especially not one containing the name of the backend.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/impl/LuceneBackendImplementor.java
```diff
@@ -1,22 +0,0 @@
...
-
-import org.hibernate.search.v6poc.backend.lucene.document.impl.LuceneRootDocumentBuilder;
-import org.hibernate.search.v6poc.backend.lucene.index.impl.IndexingBackendContext;
-import org.hibernate.search.v6poc.backend.lucene.search.query.impl.SearchBackendContext;
-import org.hibernate.search.v6poc.backend.spi.BackendImplementor;
-
-/**
- * @author Guillaume Smet
- */
-public interface LuceneBackendImplementor extends BackendImplementor<LuceneRootDocumentBuilder> {
```
gsmet: Not really fond of removing this interface. I like to have a common interface for all the different backend implementations of Lucene as we will probably have more than one.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/impl/ElasticsearchDocumentObjectBuilder.java
```diff
@@ -12,7 +12,7 @@
 import org.hibernate.search.v6poc.backend.document.DocumentElement;
 import org.hibernate.search.v6poc.backend.elasticsearch.document.model.impl.ElasticsearchIndexSchemaObjectNode;
 import org.hibernate.search.v6poc.backend.elasticsearch.gson.impl.JsonAccessor;
-import org.hibernate.search.v6poc.backend.elasticsearch.impl.MultiTenancyStrategy;
+import org.hibernate.search.v6poc.backend.elasticsearch.multitenancy.impl.MultiTenancyStrategy;
```
gsmet: Sounds sane. I wondered if it would be necessary when I did it.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -68,19 +67,19 @@ void executedRequest(String method, String path, Map<String, String> getParamete
...
 
 	@Message(id = 504, value = "Unknown field '%1$s' in indexes %2$s." )
 	SearchException unknownFieldForSearch(String absoluteFieldPath, Collection<URLEncodedString> indexNames);
 
 	@Message(id = 505, value = "Multiple conflicting types for field '%1$s': '%2$s' in index '%3$s', but '%4$s' in index '%5$s'." )
 	SearchException conflictingFieldCodecsForSearch(String absoluteFieldPath,
-			ElasticsearchFieldCodec codec1, URLEncodedString indexName1,
-			ElasticsearchFieldCodec codec2, URLEncodedString indexName2);
+			Object codec1, URLEncodedString indexName1,
+			Object codec2, URLEncodedString indexName2);
```
gsmet: Hum. Can't say I'm excited about this.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/logging/impl/Log.java
```diff
@@ -89,8 +89,8 @@
 
 	@Message(id = 511, value = "Multiple conflicting types for field '%1$s': '%2$s' in index '%3$s', but '%4$s' in index '%5$s'." )
 	SearchException conflictingFieldTypesForSearch(String absoluteFieldPath,
-			LuceneIndexSchemaFieldNode<?> schemaNode1, String indexName1,
-			LuceneIndexSchemaFieldNode<?> schemaNode2, String indexName2);
+			Object schemaNode1, String indexName1,
+			Object schemaNode2, String indexName2);
```
gsmet: Same here, not excited about this change. We usually are not very picky about the dependencies of the Log class.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchRootIndexSchemaCollectorImpl.java
```diff
@@ -9,12 +9,14 @@
...
 import org.hibernate.search.v6poc.backend.elasticsearch.document.model.impl.ElasticsearchIndexSchemaNodeCollector;
+import org.hibernate.search.v6poc.backend.elasticsearch.document.model.impl.ElasticsearchRootIndexSchemaContributor;
 import org.hibernate.search.v6poc.backend.elasticsearch.document.model.impl.esnative.RootTypeMapping;
 import org.hibernate.search.v6poc.backend.elasticsearch.document.model.impl.esnative.RoutingType;
 import org.hibernate.search.v6poc.backend.elasticsearch.multitenancy.impl.MultiTenancyStrategy;
 
 public class ElasticsearchRootIndexSchemaCollectorImpl
-		extends AbstractElasticsearchIndexSchemaCollector<IndexSchemaRootTypeNodeBuilder> {
+		extends AbstractElasticsearchIndexSchemaCollector<IndexSchemaRootTypeNodeBuilder> implements
+		ElasticsearchRootIndexSchemaContributor {
```
gsmet: I'm a bit puzzled with the class being named `Collector` and the interface being named `Contributor`. Is it expected?

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/impl/AbstractLuceneDocumentBuilder.java
```diff
@@ -36,9 +36,7 @@ protected AbstractLuceneDocumentBuilder(LuceneIndexSchemaObjectNode schemaNode)
 	}
 
 	@Override
-	public void addNestedObjectDocumentBuilder(LuceneIndexSchemaObjectNode expectedParentNode, LuceneNestedObjectDocumentBuilder nestedObjectDocumentBuilder) {
-		checkTreeConsistency( expectedParentNode );
-
+	public void addNestedObjectDocumentBuilder(LuceneNestedObjectDocumentBuilder nestedObjectDocumentBuilder) {
```
gsmet: Wondering about this one as we have the exact same pattern in Elasticsearch. Not sure what to do about it. I suppose you don't have any cyclic dependency or you would have fixed it too but I find it a bit annoying to have 2 different patterns.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/query/impl/ElasticsearchSearchTargetContext.java
```diff
@@ -25,10 +24,11 @@
 	private final SearchSortFactory<ElasticsearchSearchSortCollector> searchSortFactory;
 	private final SearchQueryFactory<ElasticsearchSearchQueryElementCollector> searchQueryFactory;
 
-	public ElasticsearchSearchTargetContext(ElasticsearchBackendImpl backend, ElasticsearchSearchTargetModel searchTargetModel) {
+	public ElasticsearchSearchTargetContext(SearchBackendContext searchWorker,
```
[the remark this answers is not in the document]
yrodiere: Right, fixed.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/query/impl/SearchBackendContext.java
```diff
@@ -0,0 +1,71 @@
...
+			MultiTenancyStrategy multiTenancyStrategy,
+			ElasticsearchWorkOrchestrator orchestrator) {
+		this.backend = backend;
+		this.multiTenancyStrategy = multiTenancyStrategy;
+		this.workFactory = workFactory;
+		this.orchestrator = orchestrator;
+
+		this.documentReferenceHitExtractor = new DocumentReferenceHitExtractor( multiTenancyStrategy );
+		this.objectHitExtractor = new ObjectHitExtractor( multiTenancyStrategy );
+		this.documentReferenceProjectionHitExtractor = new DocumentReferenceProjectionHitExtractor( multiTenancyStrategy );
```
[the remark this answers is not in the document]
yrodiere: They were previously hosted in an object that was created once per query, and there it would indeed have been a bad idea.
Here, though, the search backend context is created once per backend, and shared between all queries. So it's actually a good place to host "singletons" like this.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/query/impl/SearchQueryBuilderImpl.java
```diff
@@ -27,33 +26,37 @@
...
-	private final String tenantId;
 
-	public SearchQueryBuilderImpl(
-			ElasticsearchWorkOrchestrator queryOrchestrator,
+	SearchQueryBuilderImpl(
 			ElasticsearchWorkFactory workFactory,
-			Set<URLEncodedString> indexNames,
+			ElasticsearchWorkOrchestrator queryOrchestrator,
 			MultiTenancyStrategy multiTenancyStrategy,
-			SessionContext sessionContext,
```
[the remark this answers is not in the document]
yrodiere: It's not really an issue, it's just not necessary right now.
But we'll probably have to pass it at some point, yes. I'll revert that.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/search/query/impl/SearchBackendContext.java
```diff
@@ -0,0 +1,62 @@
...
+import org.hibernate.search.v6poc.backend.lucene.multitenancy.impl.MultiTenancyStrategy;
+import org.hibernate.search.v6poc.backend.lucene.orchestration.impl.LuceneQueryWorkOrchestrator;
+import org.hibernate.search.v6poc.backend.lucene.search.impl.LuceneSearchTargetModel;
+import org.hibernate.search.v6poc.backend.lucene.work.impl.LuceneWorkFactory;
+import org.hibernate.search.v6poc.backend.spi.BackendImplementor;
+import org.hibernate.search.v6poc.engine.spi.SessionContext;
+import org.hibernate.search.v6poc.search.query.spi.HitAggregator;
+
+public class SearchBackendContext {
+	// TODO use a dedicated object for the error context instead of the backend
```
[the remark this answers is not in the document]
yrodiere: I mentioned the problem in the section about how we should rework error handling in the google doc: https://docs.google.com/document/d/16PAa__LsxyLZcbW3q1MvgyIznh4ZnCYLupbNMXKMRds/edit?pli=1#heading=h.s83333f2qkde
This is a very broad topic, and I'm not sure how we will adress it, so for now I left it in the google doc only for discussion.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/index/impl/ElasticsearchIndexSearchTargetBuilder.java
yrodiere: Right. Done.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/impl/LuceneBackendImplementor.java
yrodiere: The backend is at the root of everything, so it depends (transitively) on almost everything. Thus, if we reference the backend from anywhere else, we will have a cyclic dependency.
Therefore we should not reference the backend from anywhere, and having an interface for backends is useless.

I can add an abstract class if you want, but even if we end up with configurable implementation bits for Lucene, I would rather move the specific bits to a separate object and use composition instead of multiple backend implementations.
Like you did for the multitenancy strategy, and like I did with the DirectoryProvider in this PR, for instance. We won't have one backend type per Directory type, we will just have a directory provider of a specific type inside the backend. If we need to make another, orthogonal concept configurable, we can then introduce another object inside the backend, whereas if we were handling these concepts in the backend directly, we would end up with a combinatorial explosion in the number of backend implementations.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/logging/impl/Log.java
yrodiere: As I wrote in the commit message, I don't care very much about dependencies of the Log class either.
But... As almost every package depends on the Log class, any dependency from the Log class to another package is very likely to create cyclic dependencies. Lots of them. And it's really hard to understand the output of the cyclic dependency analysis tool where there are a hundred cycles that you don't really care about...

So, here's the solution. I can remove it, but we'll have to do it again next time we run the analysis. And using Object here doesn't seem like a big deal, really. What's the risk? We'll probably never call this method from another place anyway.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchRootIndexSchemaCollectorImpl.java
yrodiere: Right... It's a pre-existing thing, really. The collector already had a `contribute` method before.

What is going on is this: the object collects data from the mapper through the `IndexSchemaCollector` interface, so it's a collector from the point of view of the mapper.
But the collected data must go somewhere. So this object will also contribute to the backend-specific model. Thus it is a contributor from the point of view of the backend.

Note that the same goes with field contexts: from the point of view of the mapper, they are mostly collectors, but from the point of view of the backend, they implement `ElasticsearchIndexSchemaNodeContributor`.

It's a bit convoluted, but I think makes sense. And really, there's not much choice if we want to prevent cyclic dependencies here.

Maybe we could rename the contributor interfaces to `*Model*Contributor` to make it more explicit? But then for consistency I would have to rename `ElasticsearchIndexSchemaObjectNode` and such to `*ModelNode`, and if I remember correctly you were against that.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/impl/AbstractLuceneDocumentBuilder.java
yrodiere: Ok, I'll do the same in Elasticsearch.
It's just that I didn't need to do anything in the ES integration because, there, the formatter is not responsible for adding fields, just for converting data, and the field accessor/document builder themselves take care of adding fields.

gsmet: Thanks, I think consistency is for the best in this case, even if not strictly needed.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchRootIndexSchemaCollectorImpl.java
gsmet: OK, then.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/logging/impl/Log.java
gsmet: Yeah, I can understand the nature of the problem but I'm wondering if the tools you use could be configured to ignore some packages? That would be a better fix than losing strong typing in the log methods, which is IMHO really bad.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/impl/LuceneBackendImplementor.java
gsmet: OK, sounds good.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/search/query/impl/SearchBackendContext.java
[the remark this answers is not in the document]
gsmet: OK.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/query/impl/SearchQueryBuilderImpl.java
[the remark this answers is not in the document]
gsmet: OK, cool, thanks.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/query/impl/SearchBackendContext.java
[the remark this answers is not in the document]
gsmet: Nice, thanks for the explanation.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/logging/impl/Log.java
yrodiere: I'm using IDEA's built-in tool, and there's not much we can configure.
We could setup a Sonar instance, but, well... It's way beyond the scope of this PR.

yrodiere: As we discussed, I removed the bits that annoyed you and force-pushed.  We'll use another tool to analyze cyclic dependencies in the future.
I pushed the bits I removed to https://github.com/yrodiere/hibernate-search-6-poc/tree/remove-logging-cyclic-dependencies just in case.

gsmet:
Merged, thanks!

PR #44, opened 2018-04-27
HSEARCH-1869 Validate for each hibernate.search prefixed configuration property to be consumed

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-1869

### DIFF
--- a/integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/OrmUnusedPropertiesIT.java
+++ b/integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/OrmUnusedPropertiesIT.java
@@ -1,0 +1,152 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.v6poc.integrationtest.orm;
+
+import java.util.function.Consumer;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+import org.hibernate.SessionFactory;
+import org.hibernate.boot.Metadata;
+import org.hibernate.boot.MetadataSources;
+import org.hibernate.boot.SessionFactoryBuilder;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.search.v6poc.entity.orm.cfg.SearchOrmSettings;
+import org.hibernate.search.v6poc.entity.pojo.mapping.definition.annotation.DocumentId;
+import org.hibernate.search.v6poc.entity.pojo.mapping.definition.annotation.Field;
+import org.hibernate.search.v6poc.entity.pojo.mapping.definition.annotation.Indexed;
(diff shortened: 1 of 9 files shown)


util/internal/integrationtest/common/src/main/java/org/hibernate/search/v6poc/util/impl/integrationtest/common/rule/ExpectedLog4jLog.java
```diff
@@ -0,0 +1,239 @@
+/*
```
gsmet: I think the commit message is missing `rule` after JUnit?

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/cfg/IndexingStrategyConfiguration.java
```diff
@@ -34,22 +33,22 @@
...
 	/**
-	 * Returns the {@link IndexingMode} matching the given external representation as specified via
+	 * Returns the {@link IndexingStrategyConfiguration} matching the given external representation as specified via
 	 * {@link SearchOrmSettings#INDEXING_STRATEGY}
 	 * @param indexingMode the indexing mode external representation
-	 * @return the {@link IndexingMode}
+	 * @return the {@link IndexingStrategyConfiguration}
 	 */
-	public static IndexingMode fromExternalRepresentation(String indexingMode) {
+	public static IndexingStrategyConfiguration fromExternalRepresentation(String indexingMode) {
```
gsmet: Right, I should have done it right away after my PR introducing the multi-tenancy-strategies.

integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/OrmUnusedPropertiesIT.java
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	@Test
+	public void checkEnabledByDefault_unusedProperty() {
+		String unusedPropertyKey = "hibernate.search.index.default.foo";
+		log.expectMessage(
+				"Some properties in the Hibernate Search configuration were not used",
+				"[" + unusedPropertyKey + "]"
+		);
+		log.expectMessageMissing( "Configuration property tracking is disabled" );
```
gsmet: Shouldn't we also test that a consumed property key is not logged?

util/internal/integrationtest/common/src/main/java/org/hibernate/search/v6poc/util/impl/integrationtest/common/rule/ExpectedLog4jLog.java
yrodiere: Right. Done.

integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/OrmUnusedPropertiesIT.java
yrodiere: It's already covered by the current test, since there is a single message for all properties and we constrained it to contain `"[" + unusedPropertyKey + "]"`, which means the list of properties only includes `unusedPropertyKey`.

I added this to be extra sure:

```
		// Also check that used properties are not reported as unused
		log.expectMessageMissing(
				"not used",
				DEFAULT_BACKEND_PROPERTY_KEY
		);
```

gsmet:
Merged, thanks!

PR #45, opened 2018-04-27
Try to be more consistent in the way we reference field paths/names

gsmet:


yrodiere:
Looks good; merged. Thanks!

PR #48, opened 2018-04-27
HSEARCH-3141 Search 6 groundwork - Test the model definition API in the backend TCK

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3141

### DIFF
--- a/backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaElementImpl.java
+++ b/backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaElementImpl.java
@@ -6,17 +6,23 @@
  */
 package org.hibernate.search.v6poc.backend.elasticsearch.document.model.dsl.impl;
 
+import java.lang.invoke.MethodHandles;
+
 import org.hibernate.search.v6poc.backend.document.model.dsl.IndexSchemaFieldContext;
 import org.hibernate.search.v6poc.backend.document.model.dsl.ObjectFieldStorage;
 import org.hibernate.search.v6poc.backend.document.model.dsl.spi.IndexSchemaNestingContext;
 import org.hibernate.search.v6poc.backend.elasticsearch.document.model.dsl.ElasticsearchIndexSchemaElement;
 import org.hibernate.search.v6poc.backend.elasticsearch.document.model.dsl.ElasticsearchIndexSchemaObjectField;
+import org.hibernate.search.v6poc.backend.elasticsearch.logging.impl.Log;
+import org.hibernate.search.v6poc.util.impl.common.LoggerFactory;
+import org.hibernate.search.v6poc.util.impl.common.StringHelper;
 
 /**
  * @author Yoann Rodiere
  */
 class ElasticsearchIndexSchemaElementImpl
 		implements ElasticsearchIndexSchemaElement {
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
 	protected final AbstractIndexSchemaObjectNodeBuilder nodeBuilder;
 	private final IndexSchemaNestingContext nestingContext;
@@ -39,6 +45,7 @@ class ElasticsearchIndexSchemaElementImpl
 
 	@Override
 	public IndexSchemaFieldContext field(String relativeFieldName) {
+		checkRelativeFieldName( relativeFieldName );
 		return nestingContext.nest(
 				relativeFieldName,
 				// If the field is included
@@ -56,6 +63,7 @@ class ElasticsearchIndexSchemaElementImpl
 
 	@Override
 	public ElasticsearchIndexSchemaObjectField objectField(String relativeFieldName, ObjectFieldStorage storage) {
+		checkRelativeFieldName( relativeFieldName );
 		return nestingContext.nest(
 				relativeFieldName,
 				// If the field is included
@@ -74,6 +82,15 @@ class ElasticsearchIndexSchemaElementImpl
 					nodeBuilder.setStorage( storage );
 					return new ElasticsearchIndexSchemaObjectFieldImpl( nodeBuilder, filter );
 				} );
+	}
+
+	private static void checkRelativeFieldName(String relativeFieldName) {
+		if ( StringHelper.isEmpty( relativeFieldName ) ) {
+			throw log.relativeFieldNameCannotBeNullOrEmpty( relativeFieldName );
+		}
+		if ( relativeFieldName.contains( "." ) ) {
+			throw log.relativeFieldNameCannotContainDot( relativeFieldName );
+		}
 	}
 
 }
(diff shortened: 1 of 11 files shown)


backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaElementImpl.java
```diff
@@ -76,4 +83,13 @@ public ElasticsearchIndexSchemaObjectField objectField(String relativeFieldName,
 				} );
 	}
 
+	private static void checkRelativeFieldName(String relativeFieldName) {
+		if ( relativeFieldName == null || relativeFieldName.isEmpty() ) {
```
gsmet: Minor nitpick but if we want the code to be consistent after the merge, we should probably inject Search 5 `StringHelper` and use it.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -128,4 +128,18 @@ SearchException conflictingFieldCodecsForSearch(String absoluteFieldPath,
...
+			+ " or two indexed-embeddeds may have prefixes that lead to conflicting field names,"
+			+ " or you may have declared multiple conflicting mappings."
+			+ " In any case, there is something wrong with your mapping and you should fix it." )
+	SearchException indexSchemaNodeNameConflict(String absolutePath, String name);
+
+	@Message(id = 521, value = "Field name '%1$s' is invalid: field names cannot be null or empty." )
+	SearchException relativeFieldNameCannotBeNullOrEmpty(String relativeFieldName);
+
+	@Message(id = 522, value = "Field name '%1$s' is invalid: field names cannot contain a dot ('.')."
+			+ " If you want a tree of fields, declare object field using the objectField() method." )
```
gsmet: Not sure this error message is helpful. The end user will have it while having a dot in his field name but has no clue about `objectField()`. I think we should just state that they can't use dots.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaElementImpl.java
```diff
@@ -76,4 +83,13 @@ public ElasticsearchIndexSchemaObjectField objectField(String relativeFieldName,
...
+			throw log.relativeFieldNameCannotBeNullOrEmpty( relativeFieldName );
+		}
+		if ( relativeFieldName.contains( "." ) ) {
+			throw log.relativeFieldNameCannotContainDot( relativeFieldName );
+		}
+	}
```
gsmet: Wondering if this restriction shouldn't be in the engine itself. I don't think it's a bad thing to generalize it and be consistent across backends.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/dsl/impl/LuceneIndexSchemaElementImpl.java
```diff
@@ -72,4 +79,13 @@ public LuceneIndexSchemaObjectField objectField(String relativeFieldName, Object
...
 				} );
 	}
+
+	private static void checkRelativeFieldName(String relativeFieldName) {
+		if ( relativeFieldName == null || relativeFieldName.isEmpty() ) {
+			throw log.relativeFieldNameCannotBeNullOrEmpty( relativeFieldName );
+		}
+		if ( relativeFieldName.contains( "." ) ) {
+			throw log.relativeFieldNameCannotContainDot( relativeFieldName );
+		}
```
gsmet: Obviously exactly the same comments for the Lucene part.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaElementImpl.java
yrodiere: Done

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaElementImpl.java
yrodiere: I didn't for several reasons:

1. We might want to execute this check in code where the engine is not even involved, for instance when adding a backend-specific field from a bridge.
2. We might want, for some backends, to be *more* restrictive, for instance disallow non-alphanumeric characters. I'll have to check but I doubt Elasticsearch allows just any character, there are probably some non-alphanumeric characters that are forbidden.
3. That's a bit more far-fecthed, but we might want, for some backends, to be *less* restrictive. I think Elasticsearch supports dots in field names from version 5 onward. I don't know how we would handle that when building search predicates, though.

I guess we could solve 1 & 2 by passing an object responsible for the "main" checks from the engine to the backend, and then expecting the backend to use it whenever a field is added. But I'd rather handle that in another ticket... I created [HSEARCH-3152](https://hibernate.atlassian.net/browse/HSEARCH-3152) and will do what I can when I'll address that ticket.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/dsl/impl/LuceneIndexSchemaElementImpl.java
yrodiere: Same answer

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/logging/impl/Log.java
yrodiere: The second part would have been useful to people implementing their own bridge, but as you will. I removed it.

gsmet: Well you're right in the case of field bridges, it might be useful. Maybe we should be more clear with something as:
`Field name '%1$s' is invalid: field names cannot contain a dot ('.'). Either remove the dot from your field name or if you are using a field bridge and want a tree of fields, declare object field using the objectField() method.`

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaElementImpl.java
gsmet: Keep in mind that our goal is also to provide a somewhat consistent experience across our backends.

I'm not sure it's really worth it to consider allowing special characters for one backend and not the others. I would choose the common denominator and that's it. If you can't use `§` in your field name on a backend that allows it, so be it.

And sure we can make it another issue.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/logging/impl/Log.java
yrodiere: Done!

gsmet:
Merged! Quick thought while merging in relation to this context thing we were talking: at some point, it would be nice to provide more context than just the relative field name.

PR #49, opened 2018-05-02
HSEARCH-3152 Search 6 groundwork - Implement more of the field definition DSL in the engine

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3152

This moves the implementation of `IndexSchemaElement` and `IndexSchemaObjectField`, the two classes offering entry points to define new fields in bridges, from the backend to the engine, by introducing lower-level SPIs to be implemented in backends (`IndexSchemaObjectNodeBuilder`, `IndexSchemaObjectFieldNodeBuilder` and `IndexSchemaRootNodeBuilder`).

A nice side-effect is that we no longer need `IndexSchemaCollector`, which was essentially just a `IndexSchemaElement` factory (this responsibility moved to `IndexModelBindingContext`).


backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/AbstractElasticsearchIndexSchemaObjectNodeBuilder.java
```diff
@@ -20,37 +22,44 @@
...
+	public IndexSchemaObjectFieldNodeBuilder addObjectField(String relativeFieldName, ObjectFieldStorage storage) {
+		ElasticsearchIndexSchemaObjectFieldNodeBuilder objectFieldBuilder =
+				new ElasticsearchIndexSchemaObjectFieldNodeBuilder( getAbsolutePath(), relativeFieldName, storage );
+		putProperty( relativeFieldName, objectFieldBuilder );
+		return objectFieldBuilder;
+	}
+
+	@Override
+	public IndexSchemaObjectFieldNodeBuilder createExcludedObjectField(String relativeFieldName, ObjectFieldStorage storage) {
+		return new ElasticsearchIndexSchemaObjectFieldNodeBuilder( getAbsolutePath(), relativeFieldName, storage );
```
gsmet: Thinking out loud, not sure it's doable but I was wondering if:
- we could have `createField()`/`createObjectField()` methods (which would be used for exclusion or add - but maybe you expect the exclusion code to be different at some point?)
- have `addField()`/`addObjectField()` only do the add part
- have the ^ methods return true/false depending of if successful or not
- have the log call in the engine

Not totally sure it's a good idea but wanted to share.

gsmet: It was about the log call but also about improving the symmetry between the excluded fields and the added fields.

We can improve that later though if we think we have to. I'll merge this one.

gsmet:
Merged, thanks!

PR #52, opened 2018-05-07
HSEARCH-3059 Test multi-index searches

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3059


yrodiere:
Looks good, thanks!

I'm a bit worried that we could very easily introduce a bug by adding a setting to a field type and forgetting to take it into account in `equals`/`hashCode`, though...

That's not something we could prevent with regular testing, since this new setting would use new APIs and thus wouldn't be used in tests from version N-1. Also, integration testing is probably not the right place to test that, given all the combinations of field type+settings we would have: there could easily be hundreds, maybe thousands of them.

Maybe we should consider using reflection, to systematically test all settings, and check that changing a setting leads to a codec that is considered incompatible by the backend. We could whitelist settings that we know have no impact, but the important thing is introducing a new setting could break the tests (which is good).

Anyway, that's a different problem. Merged!

PR #54, opened 2018-05-15
HSEARCH-3160 Configure an integration with SonarCloud

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3160

Results of executing all of this on my own repository with a dummy Sonar project can be seen here: https://sonarcloud.io/dashboard?id=org.hibernate.search.v6poc.sonartest%3Ahibernate-search-parent

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,19 +8,27 @@ dist: trusty
 language: java
 jdk:
   - oraclejdk8
+git:
+  # Required in order for Sonar to retrieve blame information
+  depth: false
 addons:
   apt:
     packages:
       - oracle-java8-installer
+  sonarcloud:
+    organization: "hibernate"
+    token:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -8,21 +8,45 @@ dist: trusty
...
+    -Dcoveralls.skip=true -Dsonar.skip=true
 script:
-  - ./mvnw -Pcoverage -Pjqassistant clean install javadoc:javadoc
+  # We run checks first to fail fast if there is a styling error, then we run the actual build.
+  # Note sonar must be ran in the "script:" section in order to benefit from Travis caching.
+  - ./mvnw $BUILD_OPTIONS checkstyle:check
+    && ./mvnw $BUILD_OPTIONS install javadoc:javadoc sonar:sonar -Dcheckstyle.skip
 after_success:
-  - ./mvnw coveralls:report
+  - ./mvnw $BUILD_OPTIONS coveralls:report
```
gsmet: So, with the configuration you added, Sonar doesn't throw the JaCoCo files away anymore? That was an issue when I tested.

yrodiere: It never did as far as I can tell... I checked the result in Coveralls, and everything is accounted for: https://coveralls.io/github/yrodiere/hibernate-search-6-poc
I had an issue at some point with coveralls giving me a 0% coverage, but it was because I had changed the location of `jacoco.exec` files to some place outside of the project directory.

yrodiere: Maybe that was a problem when you started the sonar plugin in another execution where you'd clean beforehand, or where you'd start JaCoCo again? Then I can understand why the files would disappear. And I can see how one could get to this situation when experimenting with the Travis configuration.

gsmet:
Rebased and merged, thanks.

.travis.yml
gsmet: Yeah, in the end, I think that was my mistake. Glad to see this one tackled.

PR #55, opened 2018-05-15
HSEARCH-3145 Rename all() predicate to matchAll()

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3145

Wanted to do that for a while but there was always ongoing PRs in the way.


yrodiere:
Merged, thanks!

PR #57, opened 2018-05-22
HSEARCH-3164 Search 6 groundwork - Do not create unnecessary identifier/routingKey bridges for @IndexedEmbeddeds

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3164


mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/processing/building/impl/PojoIndexingProcessorValueNodeBuilderDelegate.java
```diff
@@ -93,7 +92,7 @@ public void indexedEmbedded(String relativePrefix, ObjectFieldStorage storage,
 			PojoIndexingProcessorTypeNodeBuilder<V> nestedProcessorBuilder = new PojoIndexingProcessorTypeNodeBuilder<>(
 					embeddedTypeModelPath, mappingHelper, nestedBindingContext,
 					// Do NOT propagate the identity mapping collector to IndexedEmbeddeds
-					PojoIdentityMappingCollector.noOp()
+					null
```
gsmet: Just so that we end up being on the same page: considering our latest discussions and considering it's purely internal, I would have used an `Optional` here. But apparently, that's not how you see things?

yrodiere: A good article that sums up the problem: https://dzone.com/articles/optional-method-parameters

I'm a bit afraid it'll come back to bite us, but ok, considering it's internal, and bootstrap code, I'll change that. And I'll try to keep that in mind in the future.

yrodiere:
I found this issue while working on tests to check that all resources get closed. I noticed that we were instantiating too many identifier bridges and routing key bridges.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/processing/building/impl/PojoIndexingProcessorValueNodeBuilderDelegate.java
yrodiere: Done.

PR #58, opened 2018-05-22
HSEARCH-3165 Search 6 groundwork - Add a a test utils module for classes useful for any test (even unit tests)

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3165

PR #59, opened 2018-05-22
HSEARCH-3167 Exceptions may not be reported in the right order when internals use Closer.split

yrodiere:
**WARNING**: this PR is based on #58, which should be merged first.

https://hibernate.atlassian.net//browse/HSEARCH-3167

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -30,18 +30,8 @@
         </dependency>
 
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
-        </dependency>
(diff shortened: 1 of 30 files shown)


gsmet:
Merged thanks!

Wondering if the test fixes should be backported to 5? 5 is not affected by the `Closer` issue, it's something you introduced in 6?

yrodiere:
To be honest even 6 is not affected yet, we don't use `Closer#split` yet. I anticipated that at some point we'd have to close objects that throw different exceptions and wanted the method to be there just in case, because that's not something I'd want to do while working on a feature.
For now we've been lucky, and didn't need it.

PR #61, opened 2018-05-23
HSEARCH-3083 + HSEARCH-3162 Sonar violation cleanup + properly close resources on bootstrap failure

yrodiere:
~~**WARNING**: This PR is based on #57 and #60 , which should be merged first.~~ => Done

https://hibernate.atlassian.net//browse/HSEARCH-3162
https://hibernate.atlassian.net//browse/HSEARCH-3083

So... This started with a Sonar violation cleanup ([HSEARCH-3162](https://hibernate.atlassian.net//browse/HSEARCH-3162)), and ended with a broader fix for resources not being closed correctly on bootstrap failure ([HSEARCH-3083](https://hibernate.atlassian.net//browse/HSEARCH-3083)), because Sonar was reporting quite a few violations in this category.

For a full Sonar analysis of this branch, see [here](https://sonarcloud.io/dashboard?branch=sonartest-yrodiere&id=org.hibernate.search.v6poc%3Ahibernate-search-parent). To be compared with [master, here](https://sonarcloud.io/dashboard?id=org.hibernate.search.v6poc%3Ahibernate-search-parent).

PR #62, opened 2018-05-23
HSEARCH-3169 + HSEARCH-3175 Upgrade to ORM 5.3 + BeanResolver split into two

yrodiere:
~~**WARNING**: This PR based on #61 and https://github.com/hibernate/hibernate-search/pull/1680, which should be merged first.~~ => Done

https://hibernate.atlassian.net//browse/HSEARCH-3169
https://hibernate.atlassian.net/browse/HSEARCH-3175

Note: I didn't include tests for the CDI integration, because there are already tests for that in Search 5, and we will move them to 6 eventually. I just don't feel like copying the whole WildFly stuff (feature packs + integration tests) right now... Let's do that after we merged the POC into the main repository: [HSEARCH-3171](https://hibernate.atlassian.net/browse/HSEARCH-3171).


pom.xml
```diff
@@ -43,8 +43,10 @@
     </modules>
 
     <properties>
-        <version.org.hibernate>5.2.10.Final</version.org.hibernate>
+        <version.org.hibernate>5.3.0.Final</version.org.hibernate>
         <version.org.hibernate.commons.annotations>5.0.1.Final</version.org.hibernate.commons.annotations>
+        <version.javax.persistence>2.2</version.javax.persistence>
```
gsmet: +1 for following the new conventions :).

engine/src/main/java/org/hibernate/search/v6poc/engine/spi/BeanResolver.java
```diff
@@ -7,15 +7,47 @@
...
+		}
+		else if ( nameProvided ) {
+			return resolve( implementationName, expectedClass );
+		}
+		else if ( typeProvided ) {
+			return resolve( implementationType, expectedClass );
+		}
+		else {
+			throw new SearchException( "Got an empty bean reference (no name, no type)" );
+		}
```
gsmet: I haven't reviewed the 5 PR back at the time but I'm a bit worried about us having something so complex as a default method of an SPI.

yrodiere: In Search 5 We didn't, because we didn't have support for fetching beans by their name.

It's true that it's a bit complex for a default method. I can see two solutions:

1. I move this to an abstract base class, and let implementors extend from that, or not, as they see fit.
2. I add an internal layer on top of `BeanResolver`, let's say `BeanProvider`: other components would never use a `BeanResolver`, always a `BeanProvider`, and `BeanProvider` would be implemented in the engine.
`BeanResolver` would no longer expose a method accepting a `BeanReference`: just a method accepting a string reference (+ expected return type), a method accepting a type reference (+ expected return type), and a method accepting both (+ expected return type). `BeanResolver` would always expect all references to be non-null and non-empty. The third method would be optional, `BeanResolver` implementors being free to throw an exception, which would prevent users from using e.g. `@ValueBridgeBeanReference(name = "foo", type = MyType.class)`, with both attributes set. Not that currently it's always prevented, because of a limitation in how CDI integration is implemented in Hibernate ORM (CDI could handle that just fine...).
`BeanProvider` would just implement the methods currently in `BeanResolver` by delegating to the right method of a `BeanResolver`, which would include the code you referenced in your comment.

I think solution 2 would be safer in the long run, in particular because it decouples the bean resolver SPI from the other modules that use bean resolution.
Any preference?

gsmet: 2. looks goods.

What also bugged me is that an ill-designed implementation in ORM leads the default implementation of our SPI to be quite limited.

Our SPI should be neutral IMHO. And as far as I understand your proposal for 2, it would be.

yrodiere:
Done

gsmet:
Looks good to me. I added a commit to use our `StringHelper`. I let you check it's OK and merge?

Thanks!

yrodiere:
@gsmet Right, I keep forgetting about this one... Merged, thanks!

PR #63, opened 2018-05-25
HSEARCH-3177 Search 6 groundwork - Simplify the bootstrapping of mappings

yrodiere:
~~Based on #62 , which should be merged first.~~ => Done

https://hibernate.atlassian.net//browse/HSEARCH-3177


mapper-javabean/src/main/java/org/hibernate/search/v6poc/entity/javabean/JavaBeanMappingInitiator.java
```diff
@@ -9,38 +9,33 @@
...
-
-	private JavaBeanMappingInitiator(SearchMappingRepositoryBuilder mappingRepositoryBuilder,
-			JavaBeanBootstrapIntrospector introspector, boolean annotatedTypeDiscoveryEnabled, boolean multiTenancyEnabled) {
-		super(
-				mappingRepositoryBuilder, new JavaBeanMappingKey(),
-				new JavaBeanMappingFactory(),
-				introspector, false,
+		return new JavaBeanMappingInitiatorImpl(
+				mappingRepositoryBuilder, new JavaBeanBootstrapIntrospector( lookup ),
 				annotatedTypeDiscoveryEnabled, multiTenancyEnabled
```
gsmet: Is it really such a good idea to have these static methods in an API?

yrodiere: We have to offer an entry point to our API at some point, and this entry point cannot be an instance method, since users don't have any instance to call it on. So the entry point must be either a constructor or a static method.

These are the only differences I can see between a constructor and a static factory method:

* we can allow people to extend a class that exposes a public constructor, whereas classes instantiated through a static factory method cannot be easily extended.
* we can introduce logic to choose the type to instantiate in a static factory method, whereas the type is pre-determined in a constructor.

Other than that, they are practically equivalent.

I don't expect people to extend this class (if people have to extend something, I'd rather they use an SPI), so static methods are as good as constructors here in my opinion.

If you're really worried about these methods, I can make `JavaBeanMappingInitiator` a class that delegates to `JavaBeanMappingInitiatorImpl`. Then people will use the constructor of this class instead of static methods. This will require a bit of class renaming though.
Or I can introduce a  `JavaBeanMappingInitiatorFactory` class that will be instantiated through its constructor or a singleton accessor (`public static Factory get()`), then will offer non-static `create` methods.
Or, if adding more parameters is what worries you, I can use a builder pattern (`JavaBeanMappingInitiator.builder( mappingRpositoryBuilder ).enableMultiTenancy().build()`).

If I really had to pick one, I'd go with the builder solution. But I don't really feel it's necessary at the moment, and it could be added later, next to the current static methods.

gsmet: OK, let's keep it that way then.

gsmet:
Looks good to me.

yrodiere:
Merged, thanks!

PR #64, opened 2018-05-28
HSEARCH-3181 Search 6 groundwork - Port the Checkstyle/ForbiddenAPI changes from Search 5

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3181


build-config/src/main/resources/forbidden-runtime.txt
```diff
@@ -0,0 +1,36 @@
...
+# Some packages that should not be used
+java.awt.**
+sun.**
+org.slf4j.**
+junit.framework.**
+
+################################################################################################################
+# org.testng.Assert#fail() should be used instead of:
+org.assertj.core.api.Fail#fail()
+org.testng.FileAssert#fail()
```
gsmet: Still not convinced about both TestNG references (it's originally coming from HV where we are using TestNG). I complained about it but Sanne thought it was not worth fixing. I wouldn't mind having it fixed for 6 :).

build-config/src/main/resources/forbidden-tests.txt
```diff
@@ -0,0 +1,31 @@
...
+java.awt.**
+sun.**
+org.slf4j.**
+junit.framework.**
+
+################################################################################################################
+# org.testng.Assert#fail() should be used instead of:
+org.assertj.core.api.Fail#fail()
+org.testng.FileAssert#fail()
+org.assertj.core.api.Assertions#fail()
```
gsmet: Same here about the TestNG references.

build-config/src/main/resources/forbidden-runtime.txt
yrodiere: Done

build-config/src/main/resources/forbidden-tests.txt
yrodiere: And done

PR #65, opened 2018-05-29
HSEARCH-3182 Sort Hibernate ORM metadata before processing it

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3182


yrodiere:
Merged, thanks!

PR #67, opened 2018-06-08
HSEARCH-3201 Use a safe pattern to check for exception in tests

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3201

### DIFF
--- a/util/internal/test/src/main/java/org/hibernate/search/v6poc/util/impl/test/SubTest.java
+++ b/util/internal/test/src/main/java/org/hibernate/search/v6poc/util/impl/test/SubTest.java
@@ -1,0 +1,72 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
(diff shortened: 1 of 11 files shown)


util/internal/test/src/main/java/org/hibernate/search/v6poc/util/impl/test/SubTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.assertj.core.api.AbstractThrowableAssert;
+import org.assertj.core.api.Assertions;
+
+/**
+ * A util allowing to run blocks of code ("sub-tests"), expecting them to throw an exception.
+ * <p>
+ * Useful in particular when expecting an exception for each execution of a loop,
+ * in which case {@link org.junit.rules.ExpectedException} or {@link Test#expected()} cannot be used.
+ * <p>
+ * By default any throw exception will be accepted; if you want to run additional checks on the thrown exception,
```
gsmet: s/throw/thrown/

yrodiere:
Fixed the typo and merged. Thanks!

PR #68, opened 2018-06-08
HSEARCH-3197 + HSEARCH-3208 Add advanced, conditional syntax to set the minimum number of should clauses required to match for boolean predicates in the DSL

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3197
https://hibernate.atlassian.net//browse/HSEARCH-3208

This implements "minimumShouldMatch" for boolean queries: it adds support for both simple, non-conditional syntax (HSEARCH-3197) and advanced, conditional syntax (HSEARCH-3208).

This implements "minimumShouldMatch" for boolean queries only: I created [HSEARCH-3200](https://hibernate.atlassian.net/browse/HSEARCH-3200) to also implement it for match queries and simple query string queries.

**HSEARCH-3197 (at least) should be backported to Search 5, please do not mark the ticket as resolved yet.** I expect we won't be able to backport as-is, in particular because in Search 5, information is passed to the Elasticsearch backend as Lucene queries, meaning we will have to do the computation of the minimum even in the Elasticsearch case, and the Elasticsearch backend will only use the syntax where the minimum is defined as a positive integer.

### DIFF
--- a/backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
+++ b/backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
@@ -6,8 +6,15 @@
  */
 package org.hibernate.search.v6poc.backend.elasticsearch.search.predicate.impl;
 
+import java.lang.invoke.MethodHandles;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.TreeMap;
+
 import org.hibernate.search.v6poc.backend.elasticsearch.gson.impl.JsonAccessor;
+import org.hibernate.search.v6poc.backend.elasticsearch.logging.impl.Log;
 import org.hibernate.search.v6poc.search.predicate.spi.BooleanJunctionPredicateBuilder;
+import org.hibernate.search.v6poc.util.impl.common.LoggerFactory;
 
 import com.google.gson.JsonObject;
 
@@ -18,10 +25,17 @@ import com.google.gson.JsonObject;
 class BooleanJunctionPredicateBuilderImpl extends AbstractSearchPredicateBuilder
 		implements BooleanJunctionPredicateBuilder<ElasticsearchSearchPredicateCollector> {
 
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
 	private static final JsonAccessor<JsonObject> MUST = JsonAccessor.root().property( "must" ).asObject();
 	private static final JsonAccessor<JsonObject> MUST_NOT = JsonAccessor.root().property( "must_not" ).asObject();
 	private static final JsonAccessor<JsonObject> SHOULD = JsonAccessor.root().property( "should" ).asObject();
 	private static final JsonAccessor<JsonObject> FILTER = JsonAccessor.root().property( "filter" ).asObject();
+
+	private static final JsonAccessor<String> MINIMUM_SHOULD_MATCH =
+			JsonAccessor.root().property( "minimum_should_match" ).asString();
+
+	private Map<Integer, MinimumShouldMatchConstraint> minimumShouldMatchConstraints;
 
 	@Override
 	public ElasticsearchSearchPredicateCollector getMustCollector() {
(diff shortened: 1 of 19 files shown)


backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
```diff
@@ -56,6 +97,7 @@ private void mustNot(Query luceneQuery) {
 
 	private void should(Query luceneQuery) {
 		booleanQueryBuilder.add( luceneQuery, Occur.SHOULD );
+		++shouldClauseCount;
```
gsmet: First time I see the increment syntax in this order when you don't explicitly want this behavior.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
```diff
@@ -43,6 +57,34 @@ public ElasticsearchSearchPredicateCollector getFilterCollector() {
 		return this::filter;
 	}
 
+	@Override
+	public void minimumShouldMatchNumber(int constraintIgnoreCeiling, int matchingClausesNumber) {
```
gsmet: Maybe s/constraintIgnoreCeiling/applyConstraintIfHasMoreClausesThan/ .

My preference would be for `threshold` but I don't know how to convey the fact that it is strict. Maybe ask Sanne? He's generally of good advice for that.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/logging/impl/Log.java
```diff
@@ -183,4 +183,11 @@ SearchException conflictingFieldTypesForSearch(String absoluteFieldPath,
 
 	@Message(id = 543, value = "Descending order is not supported by distance sort for field '%1$s'.")
 	SearchException descendingOrderNotSupportedByDistanceSort(String absoluteFieldPath);
+
+	@Message(id = 544, value = "Computed minimum for minimumShouldMatch constraint was out of bounds:"
```
gsmet: s/was/is/?

I don't think we usually use the past tense for the log messages?

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/logging/impl/Log.java
```diff
@@ -183,4 +183,11 @@ SearchException conflictingFieldTypesForSearch(String absoluteFieldPath,
...
+			+ " expected a number between 1 and '%1$s', got '%2$s'.")
+	SearchException minimumShouldMatchMinimumOutOfBounds(int minimum, int totalShouldClauseNumber);
+
+	@Message(id = 545, value = "Multiple conflicting minimumShouldMatch constraints for ceiling '%1$s'")
+	SearchException minimumShouldMatchConflictingConstraints(int ceiling);
```
gsmet: And here we have `ceiling` again. So my suggestion for the parameter name might not be ideal.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
gsmet: Thinking about a bit more as you need a term for the log message, maybe "ignoreConstraintCeiling" would be better?

gsmet: Maybe we should ask someone native if he has a better idea on how to convey the message we want to convey.

Not sure there's a good solution as I think the issue lies in the choice of Solr to not make it a simple threshold.

gsmet:
Yeah, in the end, maybe the DSL thing is better. I like the builder approach, not sure it's consistent with the rest though.

We really need to take a decision about our syntax (lambdas or not lambdas) otherwise it's really hard to take the right path.

yrodiere:
I pushed an additional commit that changes the syntax for more complex calls (with ceilings) to something like this:

```
booleanContext6.minimumShouldMatch()
        .ifMoreThan( 4 ).thenRequirePercent( 25 )
        .ifMoreThan( 9 ).thenRequireNumber( -3 )
        .end();
```

or:

```
booleanContext6.minimumShouldMatch( c -> c
        .ifMoreThan( 4 ).thenRequirePercent( 25 )
        .ifMoreThan( 9 ).thenRequireNumber( -3 )
);
```

And yeah, we'll have to pick one syntax at some point. That's on the agenda for the next meeting.

yrodiere:
Backport PRs created:

* https://github.com/hibernate/hibernate-search/pull/1694
* https://github.com/hibernate/hibernate-search/pull/1695

gsmet:
This PR looks good to me.

yrodiere:
Merged, thanks!

PR #69, opened 2018-06-13
HSEARCH-3199 Search 6 groundwork - Restore support for fine-grained dirty detection

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3199

The goal here is to make sure that, when an entity changes, we will only trigger reindexing (of that entity or of the entities that indexed-embed it) if relevant properties changed. If only `foo` changes and `foo` is not indexed, there's no need to reindex.

Most of the code allowing to collect information about which property is used where is already here, so this PR is mainly about making the most of that information when building the reindexing resolvers, adjusting the `PojoWorker` interfaces to allow the ORM integration to tell us which properties are actually dirty, and adding tests (lots of them).

I think the dirty detection should be even finer-grained than in Search 5.

In particular, if you have an entity C indexed-embedded in A and B, and if a property of C that is used in A but not in B is dirty, Search 5 used to reindex both A and B, because it lacked the metadata to tell what it should reindex exactly.

Search 6 will only reindex A.

--- 

Don't let the number of changed lines scare you, most of them are tests:

* Changes in the main code:
> 58 files changed, 1860 insertions(+), 281 deletions(-)
* Changes in tests:
>  19 files changed, 4848 insertions(+), 1491 deletions(-)

If the documentation change is not enough to make the PR crystal-clear, maybe I can have a chat with you, @gsmet, so that I can explain the hairiest commits (and report the explanations to the technical documentation if necessary)?


orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPathFilterFactory.java
```diff
@@ -0,0 +1,290 @@
...
+		Optional<Value> valueOptional = addDirtyPathStringRepresentationsRecursively( pathsAsStrings, path, true );
+		if ( valueOptional.isPresent() ) {
+			Value value = valueOptional.get();
+			/*
+			 * We were able to resolve the path, but didn't find any Value that could possibly
+			 * be reported as dirty by Hibernate ORM.
+			 */
+			throw log.unreportedPathForDirtyChecking( persistentClass.getMappedClass(), path, value );
+		}
+		// Else everything is good, the string representation was successfully added to the set.
```
gsmet: I find the logic weird. I wouldn't have returned a path and would have throw the exception in the method below.

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPathFilterFactory.java
```diff
@@ -0,0 +1,290 @@
...
+					return Optional.empty();
+				}
+				else {
+					return Optional.of( containedValue );
+				}
+			}
+		}
+
+		throw log.unknownPathForDirtyChecking( persistentClass.getMappedClass(), path, null );
+	}
```
gsmet: Seeing this method, I'm wondering if all this complicated logic should be in ORM instead of Search?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/path/spi/PojoPathFilter.java
```diff
@@ -31,4 +29,8 @@
 	 */
 	boolean test(S paths);
 
+	static <S> PojoPathFilter<S> empty() {
+		return ignored -> false;
+	}
```
gsmet: Same question about lambdas as it was moved here.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/PojoWorker.java
```diff
@@ -15,16 +15,64 @@
...
+	 * Shorthand for {@code add(null, entity)}; see {@link #add(Object, Object)}.
+	 *
+	 * @param entity The entity to add to the index.
+	 */
 	void add(Object entity);
 
+	/**
+	 * Add an entity to the index, assuming that the entity is absent from the index.
+	 * <p>
+	 * <strong>Note:</strong> depending on the backend, this may lead to a corrupted index
```
gsmet: I would have expected potential errors but corrupted indexes?

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/event/impl/FullTextIndexEventListener.java
```diff
@@ -196,19 +201,52 @@ private void processCollectionEvent(AbstractCollectionEvent event) {
...
+		EntityPersister persister = event.getPersister();
+		final int[] dirtyProperties = event.getDirtyProperties();
+		if ( dirtyProperties != null && dirtyProperties.length > 0 ) {
+			String[] propertyNames = persister.getPropertyNames();
+			int length = dirtyProperties.length;
+			String[] dirtyPropertyNames = new String[length];
+			for ( int i = 0; i < length; i++ ) {
+				dirtyPropertyNames[i] = propertyNames[dirtyProperties[i]];
+			}
+			return dirtyPropertyNames;
```
gsmet: I find it a bit weird that you push null values for dirty properties.

integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/OrmAutomaticIndexingBasicIT.java
```diff
@@ -169,37 +169,66 @@ public void directValueUpdate_indexedElementCollectionField() {
...
+									entity1.getIndexedElementCollectionField().get( 0 )
+							)
+					)
+					.preparedThenExecuted();
+		} );
+		backendMock.verifyExpectationsMet();
+	}
+
+	/**
+	 * Test that replacing a indexed element collection
```
gsmet: s/ a / an /

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPathFilterFactory.java
```diff
@@ -0,0 +1,290 @@
...
+ *         (a dot-separated sequence of the properties mentioned in the path prefix).
+ *     </li>
+ *     <li>
+ *         Otherwise, if the whole path can be resolved to a multi-valued {@link Value}
+ *         (i.e. a {@link org.hibernate.mapping.Collection}, which also encompasses maps)
+ *         containing only embeddeds (i.e. {@link Component})
+ *         or "atomic" values (i.e. {@link SimpleValue} but not its subtypes)
+ *         or "association" values (i.e. {@link org.hibernate.mapping.OneToMany},
+ *         {@link org.hibernate.mapping.ToOne} or {@link org.hibernate.mapping.Any})
+ *         then the path will be represented as either the "property path" of that prefix
```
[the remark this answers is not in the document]
yrodiere: Copy-paste mistake. I removed "of that prefix". Same with the item above, I shouldn't have mentioned a "prefix".

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPathFilterFactory.java
```diff
@@ -0,0 +1,290 @@
...
+ *         and alternatively as the collection role.
```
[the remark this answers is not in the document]
yrodiere: Fixed:

```
 *         then the path will be represented as either the "property path"
 *         (a dot-separated sequence of the properties mentioned in the whole path).
 *         or alternatively as the collection role.
 *         <p>
 *         Note that either representation may be passed as a parameter to the filter,
 *         so we will add both representations to the set of accepted paths.
```

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPathFilterFactory.java
```diff
@@ -0,0 +1,290 @@
...
+ *         Otherwise, if a prefix of the path that does not contain any multi-valued {@link Value}
+ *         can be resolved to a Hibernate ORM {@link Value}
+ *         that will be reported by ORM as dirty in a {@link PostUpdateEvent}
+ *         and is not an association
+ *         (i.e. {@link SimpleValue} but not its subtypes),
+ *         then the path will be represented as the "property path" of that prefix.
+ *
+ *         Rationale: such values are considered as atomic when persisting,
+ *         i.e. they will be reported by ORM as dirty in a {@link PostUpdateEvent}
+ *         whenever any of their parts changed.
```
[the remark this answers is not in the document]
yrodiere: Added an example:

```java
/*         For example, {@code ContainingEntity} below will depend on the path
 *         {@code Property contained => Property field} when indexed,
 *         but any change in the property {@code field} will trigger a {@link PostUpdateEvent}
 *         on entity {@code ContainingEntity}
 *         which will allow us to tell that something in {@code contained} is dirty.
 *         In this example, the prefix leading to a value considered as "atomic when persisting"
 *         is {@code Property contained}.
 *         <pre><code>
 *             &#064;Entity
 *             &#064;Indexed
 *             public class ContainingEntity {
 *                // The user type defines an "equals(x, y)" method allowing ORM to determine whether "contained" changed
 *                &#064;Basic
 *                &#064;Type(type = "myUserType")
 *                &#064;IndexedEmbedded
 *                private Contained contained;
 *             }
 *             public class Contained {
 *                &#064;Field
 *                private String field;
 *             }
 *         </code></pre>
 */
```

... and another below:

```java
/*         For example, {@code ContainingEntity} below will depend on the path
 *         {@code Property contained => CollectionElementExtractor => Property field} when indexed,
 *         but any change in the property {@code field} will trigger an {@link org.hibernate.event.spi.AbstractCollectionEvent}
 *         which will allow us to tell that something in {@code contained} is dirty.
 *         In this example, the prefix leading to a value considered as "atomic when persisting"
 *         is {@code Property contained => CollectionElementExtractor}.
 *         <pre><code>
 *             &#064;Entity
 *             &#064;Indexed
 *             public class ContainingEntity {
 *                &#064;ElementCollection
 *                &#064;IndexedEmbedded
 *                private List&lt;Contained&gt; contained;
 *             }
 *             &#064;Embeddable
 *             public class Contained {
 *                &#064;Field
 *                private String field;
 *             }
 *         </code></pre>
 */
```

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPathFilterFactory.java
```diff
@@ -0,0 +1,290 @@
...
+	private final PersistentClass persistentClass;
+
+	public HibernateOrmPathFilterFactory(PersistentClass persistentClass) {
+		this.persistentClass = persistentClass;
+	}
+
+	@Override
+	public PojoPathFilter<Set<String>> create(Set<PojoModelPathValueNode> paths) {
+		// Use a LinkedHashSet for deterministic iteration
+		Set<String> pathsAsStrings = new LinkedHashSet<>();
```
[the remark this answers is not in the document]
yrodiere: I doubt it will change anything since most of the time it will get resized when we add a double representation for one path, leading to a result pretty much identical to an uninitialized size. But ok, done.

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPathFilterFactory.java
yrodiere: We might have two paths in some cases, so I can't return a path.

I could throw the exception in the method below, but I suspect it will make the code even more complex: I would have to perform this check (`if (isWholePath)`) everywhere we currently do `return Optional.of( value )`...

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPathFilterFactory.java
yrodiere: ORM doesn't really have a unified concept of "dirty path": sometimes it sends us a `PostUpdateEvent` with a list of dirty properties, sometimes an `AbstractCollectionEvent` with a String representing the role of the collection, ...

This code pretty much complements our `FullTextIndexEventListener`, guessing what kind of events we might receive and how the paths will be represented in each event. So I'm not sure it would make sense to push this to ORM without also pushing the listener, which sort of "unifies" events to only consider dirty paths. And I doubt that's something the ORM team will want: the event system is quite complicated, there must be a reason.

Maybe we could move *part* of the logic, such as adding `isAssociation()` getters. But walking the property tree would have to stay in Search, I think, mainly because our representation of paths is more powerful than the one in ORM: I'm not sure you can have an association wrapped in an `Optional` in ORM, for instance.

I think ORM 6 introduces a new way of visiting the metamodel, using visitors that are more advanced than what we have today. Not sure it will make the code clearer, but we can give it a try.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/path/spi/PojoPathFilter.java
yrodiere: Same answer, I fixed this one.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/PojoWorker.java
yrodiere: Ok, maybe not corrupted index, but at least wrong data in the index. Like, two documents with the same ID (in the Hibernate Search sense). The Lucene document ID would be different, though, so you're right, the index wouldn't really be "corrupted".

Changed to:

```
	 * <strong>Note:</strong> depending on the backend, this may lead to errors or duplicate entries in the index
	 * if the entity was actually already present in the index before this call.
```

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/event/impl/FullTextIndexEventListener.java
yrodiere: You're talking about the else branch below, I suppose?

You're right, I copy/pasted this from Search 5 without thinking twice.
I just checked, and `dirtyProperties` being null should not happen, ever, because of this in `org.hibernate.event.internal.DefaultFlushEntityEventListener#scheduleUpdate`:

```
		int[] dirtyProperties = event.getDirtyProperties();
		if ( event.isDirtyCheckPossible() && dirtyProperties == null ) {
			if ( !intercepted && !event.hasDirtyCollection() ) {
				throw new AssertionFailure( "dirty, but no dirty properties" );
			}
			dirtyProperties = ArrayHelper.EMPTY_INT_ARRAY;
		}
```

It might be empty, though.

I'll add a commit so that we return an empty array instead, just to avoid any NPE nightmare in the future.

integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/OrmAutomaticIndexingBasicIT.java
yrodiere: Done

yrodiere:
Merging as per our discussion. Thanks!

PR #70, opened 2018-06-14
HSEARCH-3205 Search 6 groundwork - Allow to map annotations to bridges directly, without requiring a builder

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3205

PR #71, opened 2018-06-14
HSEARCH-3078 Search 6 groundwork - Add a mapping annotation for routing key bridges

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3078

~This PR is based on #70, which should be merged first.~ => Done


gsmet:
Same here, looks good to go!

yrodiere:
Merged, thanks!

PR #72, opened 2018-06-15
HSEARCH-3073 Search 6 groundwork - Restore support for using JPA's @Id as a default for document IDs

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3073

### DIFF
--- a/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/building/impl/PojoIndexedTypeManagerBuilder.java
+++ b/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/building/impl/PojoIndexedTypeManagerBuilder.java
@@ -157,13 +157,11 @@ class PojoIndexedTypeManagerBuilder<E, D extends DocumentElement> {
 
 		@Override
 		public <T> void identifierBridge(PojoTypeModel<T> propertyTypeModel, PropertyHandle handle, IdentifierBridge<T> bridge) {
-			// FIXME ensure the bridge is closed upon build failure and when closing the SearchManagerRepository
 			this.identifierMapping = new PropertyIdentifierMapping<>( propertyTypeModel.getRawType().getCaster(), handle, bridge );
 		}
 
 		@Override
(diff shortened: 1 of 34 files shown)


mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/building/impl/PojoIndexedTypeManagerBuilder.java
```diff
@@ -140,6 +158,17 @@ void buildAndAddTo(PojoIndexedTypeManagerContainer.Builder typeManagersBuilder,
...
 	}
 
+	private Optional<BoundPojoModelPathPropertyNode<E, ?>> getEntityIdentifierPath() {
+		for ( Map.Entry<String, PojoPropertyAdditionalMetadata> entry
+				: typeAdditionalMetadata.getPropertiesAdditionalMetadata().entrySet() ) {
+			if ( entry.getValue().isEntityId() ) {
+				PojoPropertyModel<?> propertyModel = typeModel.getProperty( entry.getKey() );
+				return Optional.of( BoundPojoModelPath.root( typeModel ).property( propertyModel.getHandle() ) );
+			}
+		}
```
gsmet: We don't have a way to identify the id property when originally getting the information? It looks counter productive to iterate through the map of all the properties.

yrodiere: It's not all properties, only those that have particular additional metadata; most of the properties won't be mentioned here.

About "identify the id property when originally getting the information", I'm not sure what "when originally getting the information" is referring to exactly...

I could move the name of the entity ID to the `PojoAdditionalMetadataEntityTypeNode`. I just pushed a fix-up commit with exactly that.

Alternatively, I guess we could use a callback to retrieve such information, similarly to what we do with `PojoTypeModel`, which would enable lazy retrieval of metadata in particular. Not sure it's worth the effort for now, especially since it's SPI that we could change in 6.1.

yrodiere:
Merged, thanks

PR #73, opened 2018-06-18
Limit code duplication in ORM automatic indexing multi-valued association tests

yrodiere:
This is simply a refactoring that factorizes some of the code used in ORM automatic indexing multi-valued association tests.

These tests all do the same thing, but with a slightly different model, mainly because we have some code that is specific to each type of association in the ORM integration.

The main reason to factorize this code is that we are going to add even more test methods in future PRs, and if we don't factorize we'll have to add each test method 3 times in three different classes.

Also, we might want, in the future, to test even more different mappings; for example a `Map<String, Entity>` with `@javax.persistence.MapKey`, which derives the map key from the entity and thus might require special handling. That would require a fourth model, with a fourth test, with a fourth duplication of the test methods...


gsmet:
Looks like a nice improvement!

yrodiere:
Merged, thanks!

PR #75, opened 2018-06-20
 Limit code duplication in ORM automatic indexing association tests

yrodiere:
We already factorized code for multi-valued association tests, now let's
factorize code for tests of any association (single-valued or
multi-valued).

As a side-effect, this adds a few test methods to the multi-valued association tests that were previously missing.

### DIFF
--- a/integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/AbstractOrmAutomaticIndexingAssociationIT.java
+++ b/integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/AbstractOrmAutomaticIndexingAssociationIT.java
@@ -1,0 +1,568 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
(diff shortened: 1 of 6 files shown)


integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/AbstractOrmAutomaticIndexingAssociationIT.java
```diff
@@ -0,0 +1,568 @@
...
+public abstract class AbstractOrmAutomaticIndexingAssociationIT<
+		TIndexed extends TContaining, TContaining, TContained
+		> {
+
+	private static final String PREFIX = SearchOrmSettings.PREFIX;
+
+	@Rule
+	public BackendMock backendMock = new BackendMock( "stubBackend" );
+
+	SessionFactory sessionFactory;
```
gsmet: Raaah, again and again ;). I think you really mean `protected` here as if I have a subclass of this one in a different package, you will expect your `sessionFactory` to be visible.

yrodiere: I don't expect this class to be extended from a different package, that's why :) Anyway, it's just a test, so... I'll change it, just this once :p

yrodiere:
Merged, thanks!

PR #76, opened 2018-06-20
HSEARCH-3206 Allow to selectively disable part of the reindexing

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3206

This PR is based on #75, which should be merged first.


mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/ReindexOnUpdate.java
```diff
@@ -0,0 +1,32 @@
...
+public enum ReindexOnUpdate {
+
+	/**
+	 * Track updates to the values used in the indexing process
+	 * (or to any nested values that may be used in the indexing process)
+	 * and trigger reindexing whenever an update is detected.
+	 * <p>
+	 * Note that nested values that are assigned {@link #NO} will still be ignored.
+	 */
+	USED,
```
gsmet: Can't say I'm excited about the name. Why `USED` instead of `YES`?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
```diff
@@ -122,4 +129,32 @@ boolean isEntity(PojoRawTypeModel<?> typeModel) {
 			BoundContainerValueExtractorPath<T, V> boundExtractorPath) {
 		return extractorBinder.create( boundExtractorPath );
 	}
+
+	ReindexOnUpdate getReindexOnUpdate(ReindexOnUpdate parentReindexOnUpdate,
+			PojoTypeModel<?> typeModel, String propertyName, ContainerValueExtractorPath extractorPath) {
+		if ( ReindexOnUpdate.NO.equals( parentReindexOnUpdate ) ) {
+			return ReindexOnUpdate.NO;
+		}
```
gsmet: Wondering if we shouldn't warn the user if the parent doesn't trigger reindex but some children is marked for it.

At least I understood this condition as "if the parent doesn't trigger reindex, we ignore the settings of the children".

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoIndexingDependencyCollectorValueNode.java
```diff
@@ -59,18 +60,29 @@
...
 			PojoImplicitReindexingResolverBuildingHelper buildingHelper) {
 		super( buildingHelper );
 		this.modelPathFromRootEntityNode = modelPathFromRootEntityNode;
 		this.lastEntityNode = lastEntityNode;
 		this.modelPathFromLastEntityNode = modelPathFromLastEntityNode;
+
+		BoundPojoModelPathPropertyNode<?, P> modelPathPropertyNode = modelPathFromRootEntityNode.getParent();
+		this.reindexOnUpdate = buildingHelper.getReindexOnUpdate(
+				parentNode.getReindexOnUpdate(),
+				modelPathPropertyNode.getParent().getTypeModel(),
```
gsmet: I'm surprised you use `getParent()` here knowing you already called it once above?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/IndexingDependency.java
```diff
@@ -0,0 +1,95 @@
...
+	 * For instance, on a property of type {@code Map<EntityA, EntityB>},
+	 * {@code @AssociationInverseSide(extractors = @ContainerValueExtractorBeanReference(type = MapKeyExtractor.class))}
+	 * would define a dependency to the map keys (of type EntityA),
+	 * while {@code @AssociationInverseSide(extractors = @ContainerValueExtractorBeanReference(type = MapValueExtractor.class))}
+	 * would define a dependency to the map values (of type EntityB).
+	 * By default, Hibernate Search will try to apply a set of extractors for common types
+	 * ({@link Iterable}, {@link java.util.Collection}, {@link java.util.Optional}, ...).
+	 * To prevent Hibernate Search from applying any extractor, set this attribute to an empty array (<code>{}</code>).
+	 */
+	ContainerValueExtractorBeanReference[] extractors()
```
gsmet: I think the javadoc needs some love. The reference to `@AssociationInverseSide` is weird.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/programmatic/impl/IndexingDependencyMappingContextImpl.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * @author Yoann Rodiere
+ */
+public class IndexingDependencyMappingContextImpl
+		extends DelegatingPropertyMappingContext
+		implements IndexingDependencyMappingContext,
+		PojoMetadataContributor<PojoAdditionalMetadataCollectorPropertyNode, PojoMappingCollector> {
+
+	private ContainerValueExtractorPath extractorPath = ContainerValueExtractorPath.defaultExtractors();
+	private ReindexOnUpdate reindexOnUpdate = ReindexOnUpdate.USED;
```
yrodiere: Note to myself: this should be null, and handled accordingly in `contributeModel`.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/IndexingDependency.java
```diff
@@ -0,0 +1,95 @@
...
+	 * @return An array of reference to container value extractor implementation classes,
+	 * allowing to precisely define which value the indexing process depends on.
+	 * For instance, on a property of type {@code Map<EntityA, EntityB>},
+	 * {@code @AssociationInverseSide(extractors = @ContainerValueExtractorBeanReference(type = MapKeyExtractor.class))}
+	 * would define a dependency to the map keys (of type EntityA),
+	 * while {@code @AssociationInverseSide(extractors = @ContainerValueExtractorBeanReference(type = MapValueExtractor.class))}
+	 * would define a dependency to the map values (of type EntityB).
+	 * By default, Hibernate Search will try to apply a set of extractors for common types
+	 * ({@link Iterable}, {@link java.util.Collection}, {@link java.util.Optional}, ...).
+	 * To prevent Hibernate Search from applying any extractor, set this attribute to an empty array (<code>{}</code>).
```
yrodiere: Note to myself: this is a bad copy/paste, I should rephrase that.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/IndexingDependency.java
```diff
@@ -0,0 +1,95 @@
...
+public @interface IndexingDependency {
+
+	/**
+	 * @return How indexed entities using the annotated property should be reindexed when the value,
+	 * or any nested value, is updated.
+	 * This setting is only effective for values that are actually used when indexing
+	 * (values used in field definitions, in bridges, ...).
+	 * Defaults to {@link ReindexOnUpdate#USED}.
+	 */
+	ReindexOnUpdate reindexOnUpdate() default ReindexOnUpdate.USED;
```
gsmet: Do you expect more elements in this annotation? If not, I would not have a default value as we would expect the user to define it explicitly.

This annotation is useless when you don't change this value.

integrationtest/orm/src/test/java/org/hibernate/search/v6poc/integrationtest/orm/AbstractOrmAutomaticIndexingAssociationIT.java
```diff
@@ -559,6 +567,156 @@ public void indirectValueReplace_indexedEmbedded_elementCollectionValue_nonIndex
...
+	}
+
+	/**
+	 * Test that replacing a non-indexed, ElementCollection property in an entity
+	 * whose properties are otherwise used in an IndexedEmbedded from an indexed entity
+	 * does trigger reindexing of the indexed entity.
+	 * <p>
+	 * We need dedicated tests for this because Hibernate ORM does not handle
+	 * replaced collections the same way as it does updated collections.
+	 */
```
gsmet: I think the javadoc should be changed.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
yrodiere: You understood it right.

From what I understood, warnings that cannot be removed and are triggered by a legitimate use case are a no-no in WildFly/JBoss EAP. Maybe `INFO`?

Also, knowing that the default is to include everything, and setting something to `NO` requires explicit action from the user, I wonder if such a warning would be useful. Consider the following model:

```
@Entity
@Indexed
public class A {
    @Id
    private Integer id;
    @OneToOne
    @IndexedEmbedded
    @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.NO)
    private B b;
}
@Entity
public class B {
    @OneToOne(mappedBy = "b")
    private A a;
    @Field
    private String field;
}
```

The user explicitly said "I don't want changes to A.b to trigger reindexing of A".

When `A.b` changes, potentially the value `A.b.field` may change, but we already established that such a change should not trigger reindexing. Why should a change to `A.b.field` trigger reindexing when `A.b` didn't change, but not when `A.b` changed?

I'm honestly asking; there may be a reasoning that would make people expect that, I just don't see it.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoIndexingDependencyCollectorValueNode.java
yrodiere: I'm accessing the grandparent here. I'll change the code to make it clearer.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/IndexingDependency.java
yrodiere: I do expect more elements in the next PR, but you're right, for now the default value is useless. I'll remove it.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/IndexingDependency.java
yrodiere: Done

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/IndexingDependency.java
yrodiere: Done

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/programmatic/impl/IndexingDependencyMappingContextImpl.java
yrodiere: Done

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
gsmet: I think I wasn't clear. My issue is if the element is marked with `ReindexOnUpdate.NO` at some point but somewhere in the children, the user explicitly set it to `ReindexOnUpdate.YES`, which, if I understood correctly what you did, wouldn't change a thing.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoIndexingDependencyCollectorValueNode.java
gsmet: I confirm it's better now.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
yrodiere: So, something like this?


```
@Entity
@Indexed
public class A {
    @Id
    private Integer id;
    @OneToOne
    @IndexedEmbedded
    @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.NO)
    private B b;
}
@Entity
public class B {
    @OneToOne(mappedBy = "b")
    private A a;
    @Field
    @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.USED)
    private String field;
}
```

It would work the same as my previous example, indeed, since `USED` (`YES`, if you will) is the default.

Would you log something when there's an explicit `USED` (`YES`), but not when there's no explicit configuration?

gsmet: That was my question.

On one hand, the user might be surprised that his explicit `USED` is ignored, on the other hand, maybe he will have it set to `YES` for another hierarchy so he might be annoyed by the warning.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/ReindexOnUpdate.java
yrodiere: The rationale was that if it's unclear, at least it's not misleading, and you're more likely to actually check the javadoc, as opposed to assuming some (wrong) behavior.

Something like that:

```
@Indexed
@MyTypeBridge
public class MyEntity {
  @IndexingDependency(reindexOnChange = ReindexOnChange.YES)
  private String something;
}
```

... could very well be assumed to trigger reindexing whenever `something` changes, because it is used (without being declared) in the type bridge.

While this:

```
@Indexed
@MyTypeBridge
public class MyEntity {
  @IndexingDependency(reindexOnChange = ReindexOnChange.USED)
  private String something;
}
```

... would at least give me pause to wonder what "used" means exactly (which the javadoc could explain to me).


At least I could understand it that way. Now if you think it's far fetched, ok, let's rename it to `YES`.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
yrodiere: So... In this case, the users shouldn't even be using `reindexOnUpdate = ReindexOnUpdate.USED`, because it's the default, so this annotation is completely useless.

I could change the model, and offer a `@NoReindexOnUpdate` annotation instead, or `@IndexingDependency(ignore = true)`. That would prevent any kind of misuse. However:

1. I intended to use the same `@IndexedDependency` for defining the properties a transient property is based on; something like `@IndexedDependency(derivedFrom = {@ObjectPath(...), @ObjectPath(...)})`
2. I expected to provide, at some point in the future, an alternative `ReindexOnUpdate.SHALLOW` value, which would trigger reindexing for any value within the same entity, but never for a value in another entity that would require us to walk through associations to get to the entity to reindex.
3. I suspect that, at some point, we'll get people that want updates to all properties to be ignored, unless they explicitly tell us the opposite. In which case a value of YES could come handy.

I guess 1. could be solved by just relying on a different property for that (`@Derived(from = (derivedFrom = {@ObjectPath(...), @ObjectPath(...)})`.
2. and 3. would require to either offer one annotation per mode (`@NoReindexOnUpdate`, `@ReindexOnShallowUpdate`, `@ReindexOnUpdate`), or to use the current solution of exposing an annotation with an enum attribute.

Maybe I could simply rename `YES`/`USED` to `DEFAULT`? That would make it clear that it can't be forced, it's just the default. And later we could add other non-default modes such as `YES`/`USED` (reindex if the property is used, even if the default for the entity is `NO`) or `SHALLOW`.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/ReindexOnUpdate.java
gsmet: I think `ENABLE` / `DISABLE` would be a good compromise. It doesn't say it will do it, just that the feature is enabled.

We can probably discuss this point when we will talk about all the API we expose.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
yrodiere: Renamed `USED` to `DEFAULT` and force-pushed as per our discussion on HipChat. Will merge as soon as CI goes green.

PR #77, opened 2018-06-22
HSEARCH-1093 Smarter dirty checking with @Transient fields

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-1093

So... I ended up putting the information in the `@IndexingDependency` after all, because this is all about, well, indexing dependencies. Please have a look, then I'll see what I can do if it doesn't seem to make sense.

The first commit makes it possible to index transient properties to begin with, but only with `@IndexingDepdency(reindexOnChange = ReindexOnChange.NO)`, since we don't have any information about what this transient property is derived from.

The following commits introduce a way to tell Hibernate Search what a transient property is based on exactly.


mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/MemberPropertyHandle.java
```diff
@@ -23,15 +20,7 @@
...
-	public MemberPropertyHandle(String name, Field field) throws IllegalAccessException {
-		this( name, field, MethodHandles.lookup().unreflectGetter( field ) );
-	}
-
-	public MemberPropertyHandle(String name, Method method) throws IllegalAccessException {
-		this( name, method, MethodHandles.lookup().unreflect( method ) );
-	}
-
-	private MemberPropertyHandle(String name, Member member, MethodHandle getter) {
+	public MemberPropertyHandle(String name, Member member, MethodHandle getter) {
```
gsmet: The `MethodHandle` doesn't have a proper `toString()`, `equals()` and `hashCode()`?

If we could avoid storing the `Method`, it would be nice. From what I did on HV, it consumes a significant amount of memory.

Obviously, it only works if the `MethodHandle` provides what we are looking for for these 3 methods.

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPropertyModel.java
```diff
@@ -24,21 +24,26 @@
...
 
 	private final HibernateOrmBootstrapIntrospector introspector;
-	private final AbstractHibernateOrmTypeModel<?> holderTypeModel;
+	private final HibernateOrmRawTypeModel<?> holderTypeModel;
 
 	private final String name;
-	private final Getter getter;
+	private final Member member;
+	/**
+	 * The declared declared XProperties for this property in the holder type.
```
gsmet: s/declared declared/declared/

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPropertyModel.java
```diff
@@ -84,23 +89,31 @@ public String getName() {
...
+		if ( member instanceof Method ) {
+			return ( (Method) member ).getGenericReturnType();
 		}
-		else if ( member != null && member instanceof Field ) {
+		else if ( member instanceof Field ) {
 			return ( (Field) member ).getGenericType();
 		}
 		else {
-			return getter.getReturnType();
+			throw new AssertionFailure( "Unexpected type for a " + Member.class.getName() + ": " + member );
```
gsmet: This shouldn't happen but you don't really have the unexpected type information which is the useful one.

integrationtest/mapper-pojo/src/test/java/org/hibernate/search/v6poc/integrationtest/mapper/pojo/JavaBeanAnnotationMappingDiscoveryIT.java
```diff
@@ -162,7 +162,7 @@ public void setNonAnnotationMappedEmbedded(
 	}
 
 	/**
-	 * A type that is not registered explicitly, but mentioned in an indexed-embedded property.
+	 * DerivedFromInvalidPath type that is not registered explicitly, but mentioned in an indexed-embedded property.
```
gsmet: I suspect a refactoring issue.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
```diff
@@ -21,6 +21,7 @@
 import org.hibernate.search.v6poc.entity.pojo.extractor.impl.ContainerValueExtractorBinder;
 import org.hibernate.search.v6poc.entity.pojo.model.additionalmetadata.building.impl.PojoTypeAdditionalMetadataProvider;
 import org.hibernate.search.v6poc.entity.pojo.model.additionalmetadata.impl.PojoTypeAdditionalMetadata;
+import org.hibernate.search.v6poc.entity.pojo.model.path.PojoModelPathValueNode;
```
gsmet: s/which which/which/ in the commit message

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
```diff
@@ -157,4 +158,27 @@ ReindexOnUpdate getReindexOnUpdate(ReindexOnUpdate parentReindexOnUpdate,
...
+		Set<PojoModelPathValueNode> derivedFrom =
+				typeAdditionalMetadata.getPropertyAdditionalMetadata( propertyName )
+						.getValueAdditionalMetadata( extractorPath )
+						.getDerivedFrom();
+		if ( derivedFrom.isEmpty() ) {
+			if ( extractorBinder.isDefaultExtractorPath(
+					introspector,
+					typeModel.getProperty( propertyName ).getTypeModel(),
+					extractorPath
+			) ) {
```
gsmet: I see a lot of these things, wondering if it should be hidden somehow? I.e. here, I would have expected only one call to `getValueAdditionalMetadata()...` and I would have expected it to do the right thing for me.

Is there a special reason to not do this?

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoIndexingDependencyCollectorValueNode.java
```diff
@@ -56,27 +56,36 @@
 
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
-	private final BoundPojoModelPathValueNode<?, P, V> modelPathFromRootEntityNode;
+	private final PojoIndexingDependencyCollectorPropertyNode<?, P> parentNode;
+	private final BoundPojoModelPathValueNode<?, P, V> modelPathFromLastTypeNode;
```
gsmet: What's the difference between `modelPathFromLastEntityNode` and `modelPathFromLastTypeNode`?

I think some javadoc would help as it's pretty hard to follow. Same in the class above.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/logging/impl/Log.java
```diff
@@ -168,4 +168,18 @@ SearchException conflictingBridgeReferenceInBridgeMapping(Class<? extends Annota
 	@Message(id = 28, value = "There isn't any explicit document ID mapping for indexed type '%1$s',"
 			+ " and the entity ID cannot be used as a default because it is unknown.")
 	SearchException missingIdentifierMapping(PojoRawTypeModel<?> typeModel);
+
+	@Message(id = 29, value = "Property '%2$s' from type '%1$s' is annotated with @IndexingDependency,"
+			+ " but 'derivedFrom' contains an empty path.")
+	SearchException missingPathInIndexingDependencyDerivedFrom(PojoRawTypeModel<?> typeModel, String propertyName);
```
gsmet: I'm not sure the error message and the method name are consistent with the error. From what I understand, you could also have this error if the `ObjectPath` is just plain wrong.

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPropertyModel.java
yrodiere: Done

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/model/impl/HibernateOrmPropertyModel.java
yrodiere: Right, changed.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/MemberPropertyHandle.java
gsmet: OK, too bad...

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
gsmet: > I could hide the isDefaultExtractorPath bit inside getValueAdditionalMetadata, but then I'd have to pass more arguments than seem reasonable to that method (the container value extractor binder plus the bound path value node, in particular).

Sorry, I should have been more clear, that's what I had in mind. I think it highly depends on if we always have to consider the default extractor path case. If so, I think it has some value to hide it so that nobody has to think about it: you just pass the required arguments and you get the right info.

It might not make sense if we could have to get the `getValueAdditionalMetadata` info without considering this case.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/logging/impl/Log.java
gsmet: OK, makes sense then.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoIndexingDependencyCollectorValueNode.java
gsmet: Perfect, it's a lot clearer now.

gsmet:
@yrodiere I think it's ready to be merged!

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/dirtiness/building/impl/PojoImplicitReindexingResolverBuildingHelper.java
yrodiere: I just had a look, and it turns out we can't really move the code to `getValueAdditionalMetadata` because, in order to decide wether we have to check the default path, we have to retrieve some value and determine whether it's empty or not. In some case the value is the inverse path, in some cases it's the "derivedFrom", ... It depends.
So unless you're ok with adding some more lambdas, I think we'll have to keep it that way :)

yrodiere:
Merged, thanks!

PR #78, opened 2018-06-28
HSEARCH-3065 More work on the nested subdocuments

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3065

I'm not sure the Elasticsearch changes are totally safe as we contribute elements to the inner object, a second look wouldn't hurt.

### DIFF
--- a/backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
+++ b/backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
@@ -7,13 +7,17 @@
 package org.hibernate.search.v6poc.backend.elasticsearch.search.predicate.impl;
 
 import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.function.Consumer;
 
 import org.hibernate.search.v6poc.backend.elasticsearch.gson.impl.JsonAccessor;
 import org.hibernate.search.v6poc.backend.elasticsearch.logging.impl.Log;
 import org.hibernate.search.v6poc.search.predicate.spi.BooleanJunctionPredicateBuilder;
+import org.hibernate.search.v6poc.search.predicate.spi.SearchPredicateContributor;
 import org.hibernate.search.v6poc.util.impl.common.LoggerFactory;
 
 import com.google.gson.JsonObject;
@@ -27,6 +31,11 @@ class BooleanJunctionPredicateBuilderImpl extends AbstractSearchPredicateBuilder
 
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
+	private List<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> mustContributors;
+	private List<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> mustNotContributors;
+	private List<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> shouldContributors;
+	private List<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> filterContributors;
+
 	private static final JsonAccessor<JsonObject> MUST = JsonAccessor.root().property( "must" ).asObject();
 	private static final JsonAccessor<JsonObject> MUST_NOT = JsonAccessor.root().property( "must_not" ).asObject();
 	private static final JsonAccessor<JsonObject> SHOULD = JsonAccessor.root().property( "should" ).asObject();
@@ -38,22 +47,22 @@ class BooleanJunctionPredicateBuilderImpl extends AbstractSearchPredicateBuilder
 	private Map<Integer, MinimumShouldMatchConstraint> minimumShouldMatchConstraints;
 
 	@Override
-	public ElasticsearchSearchPredicateCollector getMustCollector() {
+	public Consumer<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> getMustCollector() {
 		return this::must;
 	}
 
 	@Override
-	public ElasticsearchSearchPredicateCollector getMustNotCollector() {
+	public Consumer<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> getMustNotCollector() {
 		return this::mustNot;
 	}
 
 	@Override
-	public ElasticsearchSearchPredicateCollector getShouldCollector() {
+	public Consumer<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> getShouldCollector() {
 		return this::should;
 	}
 
 	@Override
-	public ElasticsearchSearchPredicateCollector getFilterCollector() {
+	public Consumer<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> getFilterCollector() {
 		return this::filter;
 	}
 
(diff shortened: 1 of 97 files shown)


yrodiere:
It seems some javadoc needs to be fixed.
Reviewing now.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
```diff
@@ -38,22 +47,22 @@
 	private Map<Integer, MinimumShouldMatchConstraint> minimumShouldMatchConstraints;
 
 	@Override
-	public ElasticsearchSearchPredicateCollector getMustCollector() {
+	public Consumer<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> getMustCollector() {
```
yrodiere: You could turn this into `public void must(SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>)` or something similar, and move the weirdness of turning this to a consumer to the caller; only `org.hibernate.search.v6poc.search.dsl.predicate.impl.BooleanJunctionPredicateContextImpl.OccurContext` should need it.

It's just a suggestion though, it would make the backend code a tad simpler but it's not absolutely necessary.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/SearchPredicateFactoryImpl.java
```diff
@@ -44,65 +44,75 @@ public SearchPredicateFactoryImpl(ElasticsearchSearchTargetModel searchTargetMod
...
 	@Override
-	public SearchPredicate toSearchPredicate(SearchPredicateContributor<? super ElasticsearchSearchPredicateCollector> contributor) {
+	public Void createRootContext() {
+		return null;
+	}
+
+	@Override
+	public SearchPredicate toSearchPredicate(
+			SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector> contributor) {
+		// XXX this will be changed in a follow-up commit, passing null for now as the context
```
yrodiere: This was not changed at all in the end.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
```diff
@@ -85,25 +94,43 @@ private void addMinimumShouldMatchConstraint(int ignoreConstraintCeiling,
...
+	private void filter(SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector> contributor) {
+		if ( filterContributors == null ) {
+			filterContributors = new ArrayList<>();
+		}
+		filterContributors.add( contributor );
 	}
 
 	@Override
 	public void contribute(Void context, ElasticsearchSearchPredicateCollector collector) {
 		JsonObject innerObject = getInnerObject();
```
yrodiere: Ok, so this means, at least on the Elasticsearch side, you are re-using the same JsonObject each time `contribute()` is called.

This was not really a problem before, since `contribute()` was only supposed to be called once, but now it could theoretically be called multiple times, right? Which means we could have some issues with clauses being added twice, among others.

What do you think about adding some flag in the superclass to mark the contribute method as called, and throw an exception if called twice? Just to be safe?
We'd need to add a call to `super.contribute(context, collector)` here and in other implementations, of course.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/NestedPredicateBuilderImpl.java
```diff
@@ -23,24 +26,27 @@
...
 
+	private SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector> nestedContributor;
+
 	NestedPredicateBuilderImpl(String absoluteFieldPath) {
 		this.absoluteFieldPath = absoluteFieldPath;
 	}
 
 	@Override
-	public ElasticsearchSearchPredicateCollector getNestedCollector() {
+	public Consumer<SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>> getNestedCollector() {
```
yrodiere: Same here, you could remove this method and just promote `void nested(SearchPredicateContributor<Void, ? super ElasticsearchSearchPredicateCollector>)` to a public method.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
```diff
@@ -85,47 +87,92 @@ private void addMinimumShouldMatchConstraint(int ignoreConstraintCeiling,
...
+		if ( mustContributors != null && !mustContributors.isEmpty() ) {
+			return false;
+		}
+		if ( shouldContributors != null && !shouldContributors.isEmpty() ) {
+			return false;
+		}
+		if ( filterContributors != null && !filterContributors.isEmpty() ) {
+			return false;
+		}
+		return true;
```
yrodiere: This looks more straightforward to my eyes, but YMMV:

```
return mustNotContributors != null && !mustNotContributors.isEmpty()
         && ( mustContributors == null || mustContributors.isEmpty() )
         && ( shouldContributors == null || shouldContributors.isEmpty() )
         && ( filterContributors == null || filterContributors.isEmpty() )
```

engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/impl/MatchAllPredicateContextImpl.java
```diff
@@ -85,32 +83,25 @@ private MatchAllExceptContext getExceptContext() {
...
 
-	private class MatchAllExceptContext implements SearchPredicateDslContext<MatchAllPredicateContext<N>, CTX, C>,
-			SearchPredicateContributor<CTX, C> {
-
-		private final List<SearchPredicateContributor<CTX, ? super C>> children = new ArrayList<>();
+	private class MatchAllExceptContext implements SearchPredicateDslContext<MatchAllPredicateContext<N>, CTX, C> {
 
 		private final SearchPredicateContainerContextImpl<MatchAllPredicateContext<N>, CTX, C> containerContext;
 
+		private SearchPredicateContributor<CTX, ? super C> child;
```
yrodiere: Callers can totally write `.matchAll().except(something).except(somethingElse).except(yetAnotherThing)`.

This should be a list. Or the enclosing class should hold a `List<MatchAllExceptContext>` instead of a single `MatchAllExceptContext`; whatever your prefer. The former would allow to add multiple clauses in the lambda syntax, which is nice.
In any case, only keeping the last `except` will not work.

I pushed an additional commit to your branch with a test.

engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/impl/NestedPredicateContextImpl.java
```diff
@@ -46,7 +46,7 @@
 
 	@Override
 	public void contribute(CTX context, C collector) {
-		singlePredicateContributor.contribute( context, builder.getNestedCollector() );
+		builder.getNestedCollector().accept( singlePredicateContributor );
```
yrodiere: This should be done directly when you receive the `singlePredicateContributor`. Otherwise you run the risk of adding it twice to the builder, which is precisely what we wanted to avoid in the boolean predicate context in particular.

I know this probably works as it is in this particular case, but it's only because the consumer we're calling does not check it's only been called twice, and erases the previous value whenever it's called a second time. Two wrongs make a right, if you will :)

integrationtest/backend-tck/src/main/java/org/hibernate/search/v6poc/integrationtest/backend/tck/ObjectFieldStorageIT.java
```diff
@@ -197,7 +195,6 @@ public void search_range() {
 
 	// FIXME this test is ignored for now as the Lucene backend does not support several levels of nested
 	@Test
-	@Ignore
```
yrodiere: :tada:

integrationtest/backend-tck/src/main/java/org/hibernate/search/v6poc/integrationtest/backend/tck/NestedHierarchyIT.java
```diff
@@ -0,0 +1,351 @@
...
+import org.hibernate.search.v6poc.search.DocumentReference;
+import org.hibernate.search.v6poc.search.SearchPredicate;
+import org.hibernate.search.v6poc.search.SearchQuery;
+import org.hibernate.search.v6poc.util.impl.integrationtest.common.stub.StubSessionContext;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class NestedHierarchyIT {
```
yrodiere: Maybe rename this `NestedSearchPredicateIT`, and move it to the `org.hibernate.search.v6poc.integrationtest.backend.tck.search.predicate` package?

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/SearchPredicateFactoryImpl.java
gsmet: Right, fixed.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
gsmet: Good idea.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
gsmet: Yes, that's what worried me and why I asked you to take a closer look to this part.

Makes sense to add a safeguard. Will do.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/predicate/impl/NestedPredicateBuilderImpl.java
gsmet: Done.

integrationtest/backend-tck/src/main/java/org/hibernate/search/v6poc/integrationtest/backend/tck/NestedHierarchyIT.java
gsmet: Done.

engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/impl/NestedPredicateContextImpl.java
gsmet: RIght, done.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/search/predicate/impl/BooleanJunctionPredicateBuilderImpl.java
gsmet: Changed.

engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/impl/MatchAllPredicateContextImpl.java
gsmet: Done.

yrodiere:
Looks good. Merged, thanks!

PR #79, opened 2018-07-02
HSEARCH-3176 Use Contracts#assertNotNull()

gsmet:
https://hibernate.atlassian.net/browse/HSEARCH-3176

So gson does not depend on Guava but I added it anyway, who knows...

We have a few warnings when building the forbidden-apis configuration:
```
[WARNING] Class 'org.jboss.logging.processor.util.Objects' not found on classpath while parsing signature: org.jboss.logging.processor.util.Objects [signature ignored]
[WARNING] Class 'org.hibernate.SharedSessionContract' not found on classpath while parsing signature: org.hibernate.SharedSessionContract#getTransaction() [signature ignored]
```
and
```
[WARNING] Class 'org.jboss.logging.processor.util.Objects' not found on classpath while parsing signature: org.jboss.logging.processor.util.Objects [signature ignored]
[WARNING] Class 'org.hibernate.search.spi.SearchIntegratorBuilder' not found on classpath while parsing signature: org.hibernate.search.spi.SearchIntegratorBuilder#buildSearchIntegrator() [signature ignored]
```

As for the first one, it doesn't seem to be accessible anymore from our classpath so maybe we should remove it. If not, we should explicitly add a dependency to the build-config pom.

As for the second one, we could add an ORM dependency to the build-config pom.

As for the last one, I suppose it's coming from the old Search considering the package name? We can't really do that anyway as we would have a circular dependency.

### DIFF
--- a/build-config/src/main/resources/forbidden-runtime.txt
+++ b/build-config/src/main/resources/forbidden-runtime.txt
@@ -13,21 +13,21 @@
 java.awt.**
 sun.**
 org.slf4j.**
-junit.framework.**
+junit.framework.** @ Use the classes from org.junit, junit.framework is deprecated
 
 ################################################################################################################
 # Nobody should be using StringBuffer anymore
-java.lang.StringBuffer
+java.lang.StringBuffer @ Do not use StringBuffer: use StringBuilder
 
 ################################################################################################################
 # Probably meant the other Objects:
 
-org.jboss.logging.processor.util.Objects
+org.jboss.logging.processor.util.Objects @ Bad import, use java.util.Objects
 
 ################################################################################################################
 # Methods from Hibernate ORM :
 # This one is only safe to use with specific configurations:
-org.hibernate.SharedSessionContract#getTransaction()
+org.hibernate.SharedSessionContract#getTransaction() @ Using this method is often unsafe
 
 ################################################################################################################
 # Use our Contracts class instead
--- a/build-config/src/main/resources/forbidden-runtime.txt
+++ b/build-config/src/main/resources/forbidden-runtime.txt
@@ -28,3 +28,9 @@ org.jboss.logging.processor.util.Objects
 # Methods from Hibernate ORM :
 # This one is only safe to use with specific configurations:
 org.hibernate.SharedSessionContract#getTransaction()
+
+################################################################################################################
+# Use our Contracts class instead
+com.google.common.base.Preconditions @ Use our Contracts class instead
+java.util.Objects#requireNonNull(java.lang.Object) @ Use our Contracts class instead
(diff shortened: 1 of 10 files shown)


build-config/src/main/resources/forbidden-runtime.txt
```diff
@@ -13,18 +13,24 @@
...
 
-org.jboss.logging.processor.util.Objects
+org.jboss.logging.processor.util.Objects @ Bad import, use java.util.Objects
 
 ################################################################################################################
 # Methods from Hibernate ORM :
 # This one is only safe to use with specific configurations:
-org.hibernate.SharedSessionContract#getTransaction()
\ No newline at end of file
+org.hibernate.SharedSessionContract#getTransaction() @ Using this method is often unsafe
```
yrodiere: You could remove this comment from the parent POM now that there's a more explicit message when a violation occurs:

```
<!--
                                The method SharedSessionContract#getTransaction() should never be used
                                as it is only reliable in certain configurations.
 -->
```

yrodiere:
Looks good!

Regarding your comments in the ticket description, I'm not sure why you are suggesting to add dependencies to the build-config POM. Doesn't forbidden-apis rely on the classpath of the module being checked to find forbidden classes? Which means adding dependencies to the build-config POM would be pointless?

Regardless... I'm not a fan of adding dependencies just to silence a plugin, but if you really think it's better with a dependency, then can look the other way.

Let's keep the line about `org.hibernate.search.spi.SearchIntegratorBuilder` for now, it will make sense once again when we merge everything into Search 5.

gsmet:
Wasn't really sure about the classpath it used: the warnings saying the rules will be ignored are thrown when building the build-config module. I was wondering if it was building the rules somehow?

Or is it just because we also apply the forbidden-apis plugin to the build-config module?

build-config/src/main/resources/forbidden-runtime.txt
gsmet: Done.

yrodiere:
> Or is it just because we also apply the forbidden-apis plugin to the build-config module?

We should not, but you're right, we do. I forgot to port https://github.com/yrodiere/hibernate-search/commit/a38c7c673b9b9a7b70d34d129067db967c0045ff from Search 5.

yrodiere:
I ported https://github.com/yrodiere/hibernate-search/commit/a38c7c673b9b9a7b70d34d129067db967c0045ff from Search 5 to master, rebased to master, and the warning disappeared in the Maven logs when building the build-config module. So it's really an inspection-time thing, no need to add dependencies to allow signatures to be built correctly.

Merged, thanks!

PR #80, opened 2018-07-06
HSEARCH-3127 Allow declaring native Lucene fields in the schema

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3127

### DIFF
--- a/backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/dsl/LuceneIndexSchemaFieldContext.java
+++ b/backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/dsl/LuceneIndexSchemaFieldContext.java
@@ -18,7 +18,7 @@ import org.hibernate.search.v6poc.backend.lucene.document.model.LuceneFieldValue
 public interface LuceneIndexSchemaFieldContext extends IndexSchemaFieldContext {
 
 	/**
-	 * Declares a native field.
+	 * Declares a native field, on which projection is allowed.
 	 *
 	 * @param fieldContributor The field contributor.
 	 * @param fieldValueExtractor The field value extractor used when projecting on this field.
@@ -26,5 +26,16 @@ public interface LuceneIndexSchemaFieldContext extends IndexSchemaFieldContext {
 	 * @return The DSL context.
 	 */
 	<V> IndexSchemaFieldTerminalContext<V> asLuceneField(LuceneFieldContributor<V> fieldContributor, LuceneFieldValueExtractor<V> fieldValueExtractor);
+
+	/**
+	 * Declares a native field on which projection is not allowed.
+	 *
+	 * @param fieldContributor The field contributor.
+	 * @param <V> The type of the value.
+	 * @return The DSL context.
+	 */
+	default <V> IndexSchemaFieldTerminalContext<V> asLuceneField(LuceneFieldContributor<V> fieldContributor) {
+		return asLuceneField( fieldContributor, null );
+	}
 
 }
--- a/backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/dsl/LuceneIndexSchemaFieldContext.java
+++ b/backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/dsl/LuceneIndexSchemaFieldContext.java
@@ -6,12 +6,10 @@
  */
 package org.hibernate.search.v6poc.backend.lucene.document.model.dsl;
 
-import java.util.function.BiFunction;
-import java.util.function.Function;
-
-import org.apache.lucene.index.IndexableField;
 import org.hibernate.search.v6poc.backend.document.model.dsl.IndexSchemaFieldContext;
 import org.hibernate.search.v6poc.backend.document.model.dsl.IndexSchemaFieldTerminalContext;
+import org.hibernate.search.v6poc.backend.lucene.document.model.LuceneFieldContributor;
+import org.hibernate.search.v6poc.backend.lucene.document.model.LuceneFieldValueExtractor;
 
 
 /**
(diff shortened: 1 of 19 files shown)


backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/dsl/LuceneIndexSchemaFieldContext.java
```diff
@@ -6,12 +6,28 @@
...
 
+	/**
+	 * Declares a native field.
+	 *
+	 * @param fieldProducer A bifunction producing a field. The first parameter is the absolute path of the field, the second one is the value.
+	 * @param fieldValueExtractor A function responsible for extracting the value from the field. The parameter is the Lucene field itself.
+	 * @param <V> The type of the value.
+	 * @return The DSL context.
+	 */
+	<V> IndexSchemaFieldTerminalContext<V> asLuceneField(BiFunction<String, V, IndexableField> fieldProducer,
```
yrodiere: I expected the field to simply be of type "IndexableField", but I understand you did it this way because users are not always aware of the absolute field path, and we need to give it to them somehow.

However, as it is, the field producer can only create one `IndexableField` per value, which is a bit annoying if one wants to add DocValues, for example. Note I'm not talking about a multi-valued field (which are another matter entirely), but rather about a single value of type `V` leading to multiple `IndexableField` instances, i.e. multiple physical representations (stored, docvalues, index...).

I know the user can always create two fields with different names, and store the docvalues on one side, the indexed value on the other, but I suspect that's not how people expect to use Lucene... Especially if they've been using Hibernate Search 5 and earlier.

One option to fix this would be to use some kind of contributor/collector pattern instead:

```
public interface FieldProducer<V> {
  void produce(String absolutePath, V value, Consumer<IndexableField> collector);
}
```

Another option would be to return a specific type of `IndexFieldAccessor` for this type of fields, one that would expose the field name through a `getAbsolutePath()` method next to `write()`. Then we could make the accessor expect input values of type `IndexableField`, and the user would only have to call `write` multiple times to add multiple `IndexableField` instances to the document.

But maybe you have other ideas?

integrationtest/backend-lucene/src/test/java/org/hibernate/search/v6poc/integrationtest/backend/lucene/ExtensionIT.java
```diff
@@ -226,30 +236,115 @@ public void sort_fromJsonString_separateSort() {
 				.hasReferencesHitsExactOrder( indexName, THIRD_ID, SECOND_ID, FIRST_ID, FOURTH_ID, FIFTH_ID );
 	}
 
+	@Test
+	public void predicate_customField_throwsException() {
```
yrodiere: Maybe use the term "native" rather than "custom"? All fields are "custom" in some way, after all, even non-native ones.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/types/codec/impl/LuceneFieldFieldCodec.java
```diff
@@ -0,0 +1,70 @@
...
+		this.fieldValueExtractor = fieldValueExtractor;
+	}
+
+	@Override
+	public void encode(LuceneDocumentBuilder documentBuilder, String absoluteFieldPath, V value) {
+		if ( value == null ) {
+			return;
+		}
+
+		documentBuilder.addField( fieldProducer.apply( absoluteFieldPath, value ) );
```
yrodiere: Forgot to mention this: don't you think that, by default, we should run some sort of validation on the field name to prevent users to add `IndexableField` instances with the wrong path?
We could allow to disable those checks through a configuration property if you think it would be too expensive, but I really think checking that by default would avoid some unfortunate mistakes... Say, for example, the user could write `new StringField( "myField", value.toString() )` because they named their field `myField` and don't see why the `absoluteFieldPath` we give them would be different. And then when they use their bridge together with an `IndexedEmbedded`, KA-BOOM.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/document/model/dsl/LuceneIndexSchemaFieldContext.java
gsmet: I had an hesitation and thought about the first option too. Will do.

gsmet:
@yrodiere I think I addressed all your comments.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/types/codec/impl/LuceneFieldFieldCodec.java
```diff
@@ -36,7 +38,15 @@ public void encode(LuceneDocumentBuilder documentBuilder, String absoluteFieldPa
 			return;
 		}
 
-		fieldContributor.contribute( absoluteFieldPath, value, documentBuilder::addField );
+		List<IndexableField> fields = new ArrayList<>();
+		fieldContributor.contribute( absoluteFieldPath, value, fields::add );
```
yrodiere: You could just use a lambda that does exactly what the body of your "for" loop does. Metaspace consumption problems are the same with lambdas and method references anyway, and you would at least avoid instantiating an `ArrayList` each time the user adds a field (especially since, in this case, the list will be instantiated even if the user doesn't add anything...)

gsmet: Yup, it should have been a lambda. Forgot to commit my changes on Friday evening...

Changed and merged, thanks for the review.

PR #81, opened 2018-07-09
Backend dependency upgrades

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3229
 * https://hibernate.atlassian.net/browse/HSEARCH-3227

I copied the 5.0 init script and config to a 6.0 directory, thinking it would be better to have one per major version. But we can use the 5.0 directory if you prefer.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
         <version.org.elasticsearch.client>5.6.6</version.org.elasticsearch.client>
         <version.com.google.code.gson>2.8.0</version.com.google.code.gson>
         <version.com.google.protobuf>3.0.2</version.com.google.protobuf>
-        <version.org.apache.lucene>7.2.1</version.org.apache.lucene>
+        <version.org.apache.lucene>7.4.0</version.org.apache.lucene>
 
         <version.org.slf4j>1.7.25</version.org.slf4j>
         <version.log4j>1.2.16</version.log4j>
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
         <version.javax.persistence>2.2</version.javax.persistence>
         <version.javax.enterprise>2.0</version.javax.enterprise>
 
-        <version.org.elasticsearch.client>5.6.6</version.org.elasticsearch.client>
+        <version.org.elasticsearch.client>6.3.1</version.org.elasticsearch.client>
         <version.com.google.code.gson>2.8.0</version.com.google.code.gson>
         <version.com.google.protobuf>3.0.2</version.com.google.protobuf>
         <version.org.apache.lucene>7.4.0</version.org.apache.lucene>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -1277,9 +1277,9 @@
             </build>
         </profile>
 
-        <!-- Elasticsearch 5.2+ test environment (default) -->
+        <!-- Elasticsearch 5.6+ test environment (default) -->
```
yrodiere: This profile really does work for 5.2 and above, not just 5.6 and above. It has to do with dialects and things that changed in Elasticsearch 5.2 (normalizers). Since we did not port dialects from Search 5 yet, it's a bit confusing, granted, but still: let's keep it that way until we merge everything into the main Search repository.

pom.xml
```diff
@@ -1328,6 +1328,51 @@
             </build>
         </profile>
 
+        <!-- Elasticsearch 6.3+ test environment (default) -->
```
yrodiere: This should not be advertised as the default profile, since it is not the default profile.

util/internal/integrationtest/sharedresources/src/main/elasticsearch-maven-plugin/6.0/configuration/jvm.options
```diff
@@ -0,0 +1,110 @@
...
+# ensure the directory exists
+#-Xloggc:${loggc}
+
+# Elasticsearch 5.0.0 will throw an exception on unquoted field names in JSON.
+# If documents were already indexed with unquoted fields in a previous version
+# of Elasticsearch, some operations may throw errors.
+#
+# WARNING: This option will be removed in Elasticsearch 6.0.0 and is provided
+# only for migration purposes.
+#-Delasticsearch.json.allow_unquoted_field_names=true
```
yrodiere: It seems you didn't update the file to use the jvm.options file from Elasticsearch 6. Maybe you should? It seems a bit dodgy as it is...
Don't forget to copy the part about lowering the size of the heap space (higher in the file).

pom.xml
gsmet: Ah yes sorry, I made is so and then we discussed and changed that but forgot to update the comment.

pom.xml
gsmet: Ah yes sorry, I made is so and then we discussed and changed that but forgot to update the comment.

pom.xml
gsmet: To be honest but I didn't get to the end of it, the idea was to support latest 5.6 and 6.0 and remove everything else. But we can have this discussion another time. Will revert.

gsmet:
@yrodiere I addressed your comments and force-pushed.

yrodiere:
Tested on ES 6.0 too, it worked. Merged, thanks!

PR #82, opened 2018-07-10
HSEARCH-3231 Search 6 groundwork - Internalize index name normalization

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3231

Essentially, this makes sure that error messages and document references provided by the backend use the exact index name that was picked by the user when mapping his entities.

The internal name used by the backend may be different because of internal limitations, for example Elasticsearch is case-insensitive when it comes to index names.

It turns out there wasn't much to do on the Lucene side, since we were already using the user-provided index name as-is in documents.

Later, we may decide to add normalization/restrictions at the engine level, to fail the build when two index names are too similar (only differ by case) or use very exotic characters (non-printable characters in particular). However, this is beyond the scope of this PR, and does not even remove the need for this PR.


gsmet:
Merged, thanks!

PR #83, opened 2018-07-11
HSEARCH-2546 Give more context when there are errors while generating metadata

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2546

This PR is based on #82 , which should be merged first.

This essentially adds failure reports to the bootstrap, so that users won't have to start their application, discover an error, fix it, re-start the application to see the next error, etc. They will see a reasonably detailed failure report, with multiple failures if there are multiple failures. They may still have to start the application multiple times to fix everything, because some errors prevent Hibernate Search to continue the bootstrap and thus to find other errors, but this should be much more reasonable than it used to be.

I think this will be very, very useful to all users migrating from Search 5 in particular.

See the ticket and [the Search 6 TODO document](https://docs.google.com/document/d/16PAa__LsxyLZcbW3q1MvgyIznh4ZnCYLupbNMXKMRds/edit?pli=1#heading=h.v56txltvix6j) for discussions of this feature.

See `JavaBeanMappingFailureReportIT` for tests triggering error reports with multiple failures.


engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
```diff
@@ -279,6 +269,40 @@ void createAndAddMapper(Map<MappingKey<?>, Mapper<?>> mappers,
...
+			catch (MappingAbortedException e) {
+				ContextualFailureCollector failureCollector = buildContext.getFailureCollector();
+				Throwable cause = e.getCause();
+				if ( cause != null ) {
+					failureCollector.add( cause );
+				}
+				Throwable[] suppressed = e.getSuppressed();
+				for ( Throwable throwable : suppressed ) {
+					failureCollector.add( throwable );
+				}
```
gsmet: I don't understand the purpose of the code above going through the cause and suppressed exceptions. It seems weird to add several failures for one original failure and AFAICS you don't provide any cause to `MappingAbortedException`.

engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
```diff
@@ -279,6 +269,40 @@ void createAndAddMapper(Map<MappingKey<?>, Mapper<?>> mappers,
...
+
+				if ( !failureCollector.hasFailure() ) {
```
gsmet: I think I would have put this test at the beginning of the catch because we expect the collector to have failures from the get go, before adding the potential cause and suppressed exceptions.

integrationtest/mapper-pojo/src/test/java/org/hibernate/search/v6poc/integrationtest/mapper/pojo/JavaBeanAnnotationMappingDiscoveryIT.java
```diff
@@ -6,8 +6,6 @@
  */
 package org.hibernate.search.v6poc.integrationtest.mapper.pojo;
```
gsmet: s/definines/defines/ in the commit message

util/internal/common/src/main/java/org/hibernate/search/v6poc/util/impl/common/CollectionHelper.java
```diff
@@ -47,6 +47,13 @@ private CollectionHelper() {
 		return set;
 	}
 
+	@SafeVarargs
+	public static <T> Set<T> asLinkedHashSet(T... ts) {
+		Set<T> set = new LinkedHashSet<>( ts.length );
```
gsmet: Please use `getInitialCapacityFromExpectedSize()`. `asSet()` is incorrect and should be fixed too.

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/logging/impl/Log.java
```diff
@@ -63,4 +66,8 @@
 			+ " Hibernate ORM metadata node of type '%1$s'.")
 	SearchException invalidContainerValueExtractorForDirtyChecking(Class<?> ormMappingClass,
 			Class<? extends ContainerValueExtractor> extractorClass);
+
+	@Message(id = 10, value = "Unable to find a readable property '%2$s' on type '%1$s'")
```
gsmet: Missing dot?

mapper-javabean/src/main/java/org/hibernate/search/v6poc/entity/javabean/log/impl/Log.java
```diff
@@ -0,0 +1,30 @@
...
+import org.jboss.logging.annotations.MessageLogger;
+
+@MessageLogger(projectCode = "HSEARCH-JAVABEAN")
+public interface Log extends BasicLogger {
+
+	@Message(id = 1, value = "Unable to find property '%2$s' on type '%1$s'")
+	SearchException cannotFindProperty(@FormatWith(PojoTypeModelFormatter.class) PojoRawTypeModel<?> typeModel,
+			String propertyName);
+
+	@Message(id = 2, value = "Cannot read property '%2$s' on type '%1$s'")
```
gsmet: Until now, you were adding dots at the end of the log messages? Feel free to ignore this comment though.

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/types/dsl/impl/AbstractLuceneIndexSchemaFieldTypedContext.java
```diff
@@ -61,11 +60,6 @@ protected abstract void contribute(DeferredInitializationIndexFieldAccessor<T> r
 		return this;
 	}
 
-	@Override
-	public IndexSchemaFieldTypedContext<T> sortable(Sortable sortable) {
-		throw log.cannotUseSortableOnFieldType( relativeFieldName );
-	}
```
gsmet: I'm not sure it's totally useless. Especially if we want to remove this from the `GeoPoint` type.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaFieldContextImpl.java
```diff
@@ -7,8 +7,6 @@
 package org.hibernate.search.v6poc.backend.elasticsearch.document.model.dsl.impl;
 
 import java.time.LocalDate;
-import java.util.Arrays;
-import java.util.List;
```
gsmet: commit message: s/an manipulate lists/and manipulate lists/

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/index/impl/IndexingBackendContext.java
```diff
@@ -47,7 +46,11 @@ public IndexingBackendContext(BackendImplementor<?> backend,
 
 	@Override
 	public String toString() {
-		return getClass().getSimpleName() + "[backend=" + backend + "]";
+		return getClass().getSimpleName() + "[" + failureContext + "]";
```
gsmet: Not a big fond of it. The `failureContext` could be different in the future from what we want in the `toString()`. I don't think having a strong link between both is such a good idea.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/multitenancy/impl/NoMultiTenancyStrategyImpl.java
```diff
@@ -58,9 +58,9 @@ public String extractTenantScopedDocumentId(JsonObject hit) {
 	}
 
 	@Override
-	public void checkTenantId(BackendImplementor<?> backend, String tenantId) {
+	public void checkTenantId(String tenantId, FailureContext backendContext) {
 		if ( tenantId != null ) {
-			throw log.tenantIdProvidedButMultiTenancyDisabled( backend, tenantId );
+			throw log.tenantIdProvidedButMultiTenancyDisabled( tenantId, backendContext );
```
gsmet: Maybe `backendFailureContext`.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/query/impl/SearchBackendContext.java
```diff
@@ -51,7 +50,11 @@ public SearchBackendContext(BackendImplementor<?> backend,
 
 	@Override
 	public String toString() {
-		return getClass().getSimpleName() + "[backend=" + backend + "]";
+		return getClass().getSimpleName() + "[" + failureContext + "]";
```
gsmet: Same here about linking `toString()` and failure contexts.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -71,36 +70,38 @@ void executedRequest(String method, String path, Map<String, String> getParamete
...
+	SearchExceptionWithContext unknownFieldForSearch(String absoluteFieldPath, @Param FailureContext context);
 
-	@Message(id = 505, value = "Multiple conflicting types for field '%1$s': '%2$s' in index '%3$s', but '%4$s' in index '%5$s'." )
-	SearchException conflictingFieldTypesForSearch(String absoluteFieldPath,
-			ElasticsearchIndexSchemaFieldNode schemaNode1, String indexName1,
-			ElasticsearchIndexSchemaFieldNode schemaNode2, String indexName2);
+	@Message(id = 505, value = "Multiple conflicting types for field '%1$s': '%2$s' vs. '%3$s'." )
+	SearchExceptionWithContext conflictingFieldTypesForSearch(String absoluteFieldPath,
+			ElasticsearchIndexSchemaFieldNode schemaNode1, ElasticsearchIndexSchemaFieldNode schemaNode2,
+			@Param FailureContext context);
```
gsmet: Not sure it's an improvement here. You lose the link between the node and the index. But maybe it's not a big deal if it brings consistency.

engine/src/main/java/org/hibernate/search/v6poc/logging/impl/RootFailureCollector.java
```diff
@@ -0,0 +1,177 @@
...
+			throw log.bootstrapCollectedFailures( renderedFailures );
+		}
+	}
+
+	private String renderFailures() {
+		ToStringStyle style = ToStringStyle.multilineIndentStructure(
+				MESSAGES.contextFailuresSeparator(),
+				MESSAGES.contextIndent(),
+				MESSAGES.contextFailuresBulletPoint(),
+				MESSAGES.contextFailuresNoBulletPoint()
```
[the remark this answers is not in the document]
yrodiere: I didn't know if languages such as Japanese, Corean, Arabic, Chinese use the same symbols as we do for bullet points and colons. Also didn't know whether the chosen separators would work well with right-to-left languages. So I just took the safest path.
It will only ever happen on failure anyway, and it's just a few strings. Not a big deal.

engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
yrodiere: `MappingAbortedException` is SPI, so theoretically someone else could be throwing it, but mainly it's to be safe in case we add a cause/suppressed exception to `MappingAbortedException` in the future. Missing exceptions in a stack trace are the worst.
I added a comment to clarify that.

engine/src/main/java/org/hibernate/search/v6poc/engine/impl/SearchMappingRepositoryBuilderImpl.java
yrodiere: Right, done.

integrationtest/mapper-pojo/src/test/java/org/hibernate/search/v6poc/integrationtest/mapper/pojo/JavaBeanAnnotationMappingDiscoveryIT.java
yrodiere: Done

mapper-javabean/src/main/java/org/hibernate/search/v6poc/entity/javabean/log/impl/Log.java
yrodiere: Done

orm/src/main/java/org/hibernate/search/v6poc/entity/orm/logging/impl/Log.java
yrodiere: Done

util/internal/common/src/main/java/org/hibernate/search/v6poc/util/impl/common/CollectionHelper.java
yrodiere: Done

backend-lucene/src/main/java/org/hibernate/search/v6poc/backend/lucene/types/dsl/impl/AbstractLuceneIndexSchemaFieldTypedContext.java
yrodiere: It was overridden in every subclass that actually exposed it, so it was useless and I was unable to test it. But we can add it back later, sure.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaFieldContextImpl.java
yrodiere: Done

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/index/impl/IndexingBackendContext.java
yrodiere: Are you sure it's really important? I would expect the "failure context" used in the "Indexing backend context" to translate more or less accurately what we want in the `toString()` method... I mean, if it's good enough for users, it should be good enough for us.
The alternative is to keep a reference to a `BackendImplementor` object whose methods we don't use, which is really weird.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/logging/impl/Log.java
yrodiere: Yeah, I don't think it's very important since, in order to fix this error, you will very probably have to have a look at both indexes anyway.
Besides, we still display the nodes in the same order as index names, so even if it's not very clear, users could guess which node is in which index, and they would guess right.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/multitenancy/impl/NoMultiTenancyStrategyImpl.java
yrodiere: Given the next commits that rename it to `EventContext`, not sure it's necessary? `EventContext backendContext` seems to make sense to me.

backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/search/query/impl/SearchBackendContext.java
yrodiere: Same answer, see above.

mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/mapping/definition/annotation/impl/TypeAnnotationProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+		processEach( mappingContext, typeModel, extractAnnotations( typeModel ) );
+	}
+
+	abstract Stream<? extends A> extractAnnotations(PojoRawTypeModel<?> typeModel);
+
+	abstract void doProcess(TypeMappingContext mappingContext, PojoRawTypeModel<?> typeModel, A annotation);
+
+	private void processEach(TypeMappingContext mappingContext, PojoRawTypeModel<?> typeModel,
+			Stream<? extends A> annotations) {
+		List<A> annotationList = annotations.collect( Collectors.toList() );
```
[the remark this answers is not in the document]
gsmet: Well, I was just saying that if we go the "let's remove the lambdas" route, streams will probably have to go as they are only interesting if you can use lambdas and method references.

gsmet:
@yrodiere looks good to merge, I'll add the `sortable()` method again in a later patch.

yrodiere:
Rebased on master and merged. Thanks!

PR #84, opened 2018-07-17
HSEARCH-3220 Stop using the contributor/collector pattern in the predicate/sort builders

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3220

* The engine code related to the query DSL is now a bit clearer, with a clear separation between the contributors from the DSL and the builders from the backend.
* The concept of "context" (for nested fields) is now buried inside the Lucene backend, not visible anywhere in the engine or in the Elasticsearch backend.
* The predicate/sort DSL implementations in backends are slightly clearer, since they now only deal with their own builder interfaces, not with a "contributor" interface from the engine. Most of the code works the same, though.
* I managed to still only have one generic type parameter for the query builder, which should allow us to add as many concepts as we want (aggregations, facets, ...) without ending up with a billion generic type parameters in the `SearchQueryBuilder` interface.

That's a lot of changes, but let's look at it on the bright side: the whole thing should be transparent to users, since I didn't have to change any API.

Maybe the best way to review it would be to just have a look at the end result (use the "split" view for diffs and just look at the right part).


integrationtest/backend-elasticsearch/src/test/java/org/hibernate/search/v6poc/integrationtest/backend/elasticsearch/util/ElasticsearchTckTestRunner.java
```diff
@@ -22,6 +22,6 @@
  * @author Yoann Rodiere
  */
 @RunWith(ClasspathSuite.class)
-@ClasspathSuite.ClassnameFilters({ ".*\\.tck\\..*" })
+@ClasspathSuite.ClassnameFilters({ ".*\\.tck\\..*SearchSortIT.*" })
```
gsmet: I don't think you want that.

yrodiere:
Nice catch! Fixed and merged, thanks.

PR #85, opened 2018-07-19
HSEARCH-3240 Use the pre-installed JDK in the Travis build instead of trying to update it in each build

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3240

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,6 @@ git:
   # Required in order for Sonar to retrieve blame information
   depth: false
 addons:
-  apt:
-    packages:
-      - oracle-java8-installer
   sonarcloud:
     organization: "hibernate"
     token:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,7 +3,7 @@
 # therefore deactivated. Activating Travis for your own fork is as easy as
 # activating it in the travis site of your fork.
 
-sudo: required
+sudo: false
```
gsmet: I don't know if it's still true but at the time we also used that to use the new containerized infra.

Maybe they are all run on the same infra, now.

yrodiere: I don't know about earlier, but currently, according to [this](https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System), `sudo: required` means we *don't* use containers, but VMs.

yrodiere:
Merged following the discussion on https://github.com/hibernate/hibernate-search/pull/1710

PR #87, opened 2018-07-19
HSEARCH-3241 Reorganize and clarify Maven properties used to pass JVM args

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3241

Should only be merged after https://github.com/hibernate/hibernate-search/pull/1711


yrodiere:
Merged, thanks!

PR #91, opened 2018-07-31
HSEARCH-3191 Use .end() terminal calls for all predicates

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3191

AFAICS, the sorts are already covered.


engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/impl/MatchPredicateFieldSetContextImpl.java
```diff
@@ -71,12 +72,17 @@ public void contributePredicateBuilders(Consumer<B> collector) {
...
 		}
 
-		public N matching(Object value) {
+		public ExplicitEndContext<N> matching(Object value) {
 			if ( value == null ) {
 				throw log.matchPredicateCannotMatchNullValue( collectAbsoluteFieldPaths() );
 			}
 			getQueryBuilders().forEach( b -> b.value( value ) );
-			return getNextContextProvider().get();
+			return new ExplicitEndContext<N>() {
```
yrodiere: One less object is not a big deal, I know, but you could just make `MatchPredicateFieldSetContextImpl` implement `ExplicitEndContext` and return `this` here. Like we did in other places.
Or, even better, you could make `MultiFieldPredicateCommonState` implement `ExplicitEndContext`, and just return the common state here.

engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/MatchPredicateFieldSetContext.java
```diff
@@ -14,6 +15,6 @@
  */
 public interface MatchPredicateFieldSetContext<N> extends MultiFieldPredicateFieldSetContext<MatchPredicateFieldSetContext<N>> {
 
-	N matching(Object value);
+	ExplicitEndContext<N> matching(Object value);
```
yrodiere: So, in the spatial DSL interfaces (`SpatialWithinPredicateFieldSetContext` in particular), we return specific interfaces for each method: `SpatialWithinBoundingBoxPredicateContext`, `SpatialWithinCirclePredicateContext` and `SpatialWithinPolygonPredicateContext`. Here we just return the generic `ExplicitEndContext`. It feels a bit inconsistent.

In my opinion, what you did here is better. If we need to add more methods one day, we will be able to easily replace `ExplicitEndContext` in the signature of this method with a subinterface.

Don't you think we should do the same for the spatial DSL interfaces?

Note that, strangely, the "end" context interfaces in the spatial DSL do not extend `ExplicitEndContext`, even though they define an `end()` method.

engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/impl/RangePredicateFieldSetContextImpl.java
```diff
@@ -125,10 +126,15 @@ N above(Object value, RangeBoundInclusion inclusion) {
 					break;
 			}
 			checkHasNonNullBound();
-			return getNextContextProvider().get();
+			return new ExplicitEndContext<N>() {
```
yrodiere: Same as above: you could just make `RangePredicateFieldSetContextImpl` implement `ExplicitEndContext` and return `this` here. Like we did in other places.
Or, even better, you could make `MultiFieldPredicateCommonState` implement `ExplicitEndContext`, and just return the common state here.

engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/impl/RangePredicateFieldSetContextImpl.java
```diff
@@ -141,7 +147,12 @@ N below(Object value, RangeBoundInclusion inclusion) {
...
```
yrodiere: Same as above: you could just make `RangePredicateFieldSetContextImpl` implement `ExplicitEndContext` and return `this` here. Like we did in other places.
Or, even better, you could make `MultiFieldPredicateCommonState` implement `ExplicitEndContext`, and just return the common state here.

PR #97, opened 2018-08-31
Follow-up: HSEARCH-3294 Send email notifications upon build failures in the Jenkinsfile

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3294

Adding a way to configure which email addresses are used as notification recipients.

This is necessary because automatic email address detection from git changesets is not as reliable as I thought (I think it fails for the first build, or when force-pushing). When building branches in forks, we used to *only* rely on this automatic email address detection, because we didn't want to send notifications to, for example, Guillaume, when a branch in my fork fails to build.

So if we want to get proper notifications for branch builds in our forks, if , we must have a way to configure the notification recipients per fork. This is what this commit adds.

Failure to provide a configuration file will not fail the build, so this does not impair the ability to use the Jenkinsfile "out of the box", without configuration: http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-6-poc-yoann/detail/HSEARCH-3294/15/pipeline/4

Adding the correct configuration file to the job will, as expected, add additional recipients to email notifications: http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-6-poc-yoann/detail/HSEARCH-3294/16/pipeline/4

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -50,6 +50,7 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
  * - new java.lang.IllegalArgumentException java.lang.String
  * - method hudson.plugins.git.GitSCM getUserRemoteConfigs
  * - method hudson.plugins.git.UserRemoteConfig getUrl
+ * - method java.lang.Throwable addSuppressed java.lang.Throwable
  *
  * Just run the script a few times, it will fail and display a link to allow these calls.
  */
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,9 +6,8 @@
  */
 
 import groovy.transform.Field
+import groovy.json.JsonSlurper
 import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
-
-import java.util.regex.Pattern
 
 /*
  * WARNING: DO NOT IMPORT LOCAL LIBRARIES HERE.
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -34,25 +41,29 @@ import java.util.regex.Pattern
...
-@Field final Pattern MAIN_REPOSITORY_URL_PATTERN = ~/^(git@github.com:|https:\/\/github.com\/)hibernate\/.*.git$/
-// The following should be a space-separated list of email addresses
-@Field final String MAIN_REPOSITORY_MAINTAINER_EMAILS = "yoann+hibernate-ci@hibernate.org guillaume+ci@hibernate.org"
-
+// Default node pattern, to be used for resource-intensive stages.
+// Should not include the master node.
 @Field final String NODE_PATTERN_BASE = 'Slave'
+// Quick-use node pattern, to be used for very light, quick, and environment-independent stages,
+// such as sending a notification. May include the master node in particular.
+@Field final String QUICK_USE_NODE_PATTERN = 'Master||Slave'
```
gsmet: I don't understand why we would need that? Shouldn't the configuration file parsing be done on the master before distributing the job to the slaves?

yrodiere: Previously the configuration stage did not have a `node` statement, meaning everything indeed happened on the master node, but some features were not available. In particular, filesystem access is disabled. You have to allocate a node explicitly for that. Which is why I added a `node()` statement just to fetch the configuration file.

I'd rather not require to use a slave node for that, since they might all be busy, and it would be a shame to wait for a node just to access a configuration file.

As long as we don't use the master node for anything else than very short operations, that should not slow down the build (it just takes one or two seconds).

You can try without the `node` statement if you want, you'll just get an exception telling you that some context is missing.

gsmet: Well, what's bothering is more than we could execute this on a slave. I think I would have made it master only.

Anyway, we can merge it.

yrodiere: It should work on a slave as well. The configuration files are accessible on any node, that's the point of this plugin.
Merging, thanks.

PR #99, opened 2018-08-31
HSEARCH-3196 Give up on per-module project codes and book ranges of IDs per module instead

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3196

As mentioned in the commit message:

```
Unfortunately, this means I had to change the ID of many, many messages.
Since this will make git blame unusable for these files anyway, I
decided to take advantage of this and make the syntax of all the
@message annotations consistent.
```

In particular I took care to put message strings on a different line than message IDs, so that we won't mess up the git history for the lines containing message string when we change message IDs (though I'd rather not do that again...).

Note that this does not add more changes to this patch, because I had to change these lines anyway (to change the message ID).

Not sure I'm making sense... Let's talk about it on HipChat if I'm unclear.

### DIFF
--- a/util/internal/common/src/main/java/org/hibernate/search/v6poc/util/impl/common/MessageConstants.java
+++ b/util/internal/common/src/main/java/org/hibernate/search/v6poc/util/impl/common/MessageConstants.java
@@ -1,0 +1,67 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.v6poc.util.impl.common;
+
+public final class MessageConstants {
+
+	private MessageConstants() {
+	}
+
+	public static final String PROJECT_CODE = "HSEARCH";
+
+	// -----------------------------------
+	// Message ID ranges
+	// These ID ranges should NOT OVERLAP
+	// -----------------------------------
+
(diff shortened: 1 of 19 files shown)


util/internal/common/src/main/java/org/hibernate/search/v6poc/util/impl/common/MessageConstants.java
```diff
@@ -0,0 +1,67 @@
...
+	public static final int BACKEND_LUCENE_ID_RANGE_MAX = 609999;
+
+	public static final int MAPPER_POJO_ID_RANGE_MIN = 700000;
+	public static final int MAPPER_POJO_ID_RANGE_MAX = 700999;
+
+	public static final int MAPPER_JAVABEAN_ID_RANGE_MIN = 750000;
+	public static final int MAPPER_JAVABEAN_ID_RANGE_MAX = 759999;
+
+	public static final int ORM_ID_RANGE_MIN = 800000;
+	public static final int ORM_ID_RANGE_MAX = 800999;
```
gsmet: For mapper or orm, you defined the range for only 999 messages. I would have made the ranges equivalent (even if we don't expect that many messages).

It's even more strange for these two:
* MAPPER_POJO -> 999 messages
* MAPPER_JAVABEAN -> 9999 messages

yrodiere: Right. I just raised the maximums to allow 10000 messages everywhere.

yrodiere:
Merged, thanks!

PR #101, opened 2018-09-04
HSEARCH-3139 Avoid SPIs leaking into APIs

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3139

Best reviewed commit by commit.


engine/src/main/java/org/hibernate/search/v6poc/search/dsl/predicate/SearchPredicateContainerContextExtension.java
```diff
@@ -27,6 +33,8 @@
 	/**
 	 * Attempt to extend a given context, throwing an exception in case of failure.
 	 *
+	 * <strong>WARNING:</strong> this method is not API, see comments at the type level.
```
gsmet: I think a `<p>` before this sentence would be in order, WDYT?

Same for the others below.

yrodiere: Right. Done.

yrodiere:
Applied your suggestion and merged. Thanks!

